PR #82, opened 2020-07-06
Add drop commands to copy config and examples

kburns:
This PR modifies the setup.py build to tar up any .py files in the examples folder, and adds command-line commands for producing local copies of the config file as well as the example tree.

### DIFF
--- a/dedalus/__main__.py
+++ b/dedalus/__main__.py
@@ -5,8 +5,8 @@ Usage:
     dedalus test
     dedalus bench
     dedalus cov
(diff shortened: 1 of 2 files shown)


dedalus/__main__.py
```diff
@@ -27,6 +32,13 @@
         bench()
     elif args['cov']:
         cov()
+    elif args['drop_config']:
```
jsoishi, line 35: make sure to change here too...

PR #92, opened 2020-10-01
Speedup arithmetic using numexpr

kburns:
This PR starts with Daniel's work on reimplementing some arithmetic operators using better plain python, cython, and numexpr. Given that numexpr seems to work well on multiple systems and is not a heavy dependency, I think it makes sense to just reimplement everything using numexpr, since it's much cleaner than the other implementations.

Running to-do list for getting this done:
- [x] Merge in blas-based timestepper optimizations from d3.
- [x] Simplify current changes down to just numexpr implementations.
- [x] Implement fast `curl` for more than just spherical coordinates.
- [x] Fix `multiply`: current implementation fails for outer products of tensors.
- [x] Implement numexpr-versions of spin recombinations.
- [ ] Check real and complex


kburns:
@lecoanet, unit tests on my laptop seem to indicate that numexpr is about 10x slower than `np.multiply` and `np.add` for simple array-number and array-array multiplication and addition.  Did you see numexpr being faster for these operations?  If not maybe we should keep the numpy operations in the simple cases, and only use numexpr for things that require multiple operations and temporaries.

kburns:
Tried a new cross-product approach: just assign coordsystems a handedness, then pick left or right-handed CrossProduct operate methods at instantiation.

lecoanet:
I think numexpr was faster than np.multiply and np.add on Pleiades. But I'll check again and verify. I think these were ~1-2% of the simulation time when I was using numpy, so there's no way they're 10x slower with numexpr on Pleiades, or we would be spending a huge about of time on them.

kburns:
Ok looking closer, the 10x gap was for pretty small arrays (32 x 32) on my laptop.  For more realistic sizes (512 x 32 x 32), I still see about a factor of 2x better performance for numpy on my laptop and the flatiron cluster.

lecoanet:
Ok, I just tested using numpy vs numexpr for `AddFields`, `MultiplyFields`, and `MultiplyNumberFields` on Pleiades. I ran our fiducial shell convection problem with resolution Nmax=511, Lmax=110 on 1024 cores. Here are the timing results:

commit 58b047d; run yesterday:
Total time: 299.25987792015076
Arithmetic 225 (AddFields): 0.64% = 1.915
Arithmetic 584 (DotProduct): 0.37% = 1.107
Arithmetic 639 (CrossProduct): 0.59% = 1.766
Arithmetic 746 (MultiplyFields): 0.32% = 0.958
Arithmetic 798 (MultiplyNumberField): 0.1% = 0.299

commit ccb90ad from my fork; run yesterday:
Total time: 299.0743188858032
Arithmetic 230 (AddFields): 0.64% = 1.914
Arithmetic 605 (DotProduct): 0.37% = 1.107
Arithmetic 658 (CrossProduct): 0.60% = 1.794
Arithmetic 761 (MultiplyFields): 0.33% = 0.987
Arithmetic 850 (MultiplyNumberField): 0.1% = 0.299

commit 58b047d, but switched to numpy, run today:
Total time: 222.15662622451782
Arithmetic 225 (AddFields): 0.96% = 2.133
Arithmetic 585 (DotProduct): 0.58% = 1.1289
Arithmetic 640 (CrossProduct): 0.93% = 2.066
Arithmetic 747 (MultiplyFields): 0.5% = 1.111
Arithmetic 800 (MultiplyNumberField): 0.12% = 0.267

commit ccb90ad from my fork; run today:
Total time: 222.15398287773132
Arithmetic 230 (AddFields): 0.97% = 2.155
Arithmetic 605 (DotProduct): 0.56% = 1.244
Arithmetic 658 (CrossProduct): 0.91% = 2.022
Arithmetic 761 (MultiplyFields): 0.51% = 1.133
Arithmetic 850 (MultiplyNumberField): 0.15% = 0.333

When I switched to numpy I used the following for `AddFields`, `MultiplyFields`, and `MultiplyNumberFields`, respectively:

> np.add(args[0].data, args[1].data, out=out.data)
> np.multiply(args[0].data, args[1].data, out=out.data)
> np.multiply(arg0, arg1.data, out=out.data)

For whatever reason, Pleiades was much faster this morning than last night (299 seconds -> 222 seconds). However, the arithmetic was slower this morning than last night! Overall, there doesn't seem to be any speed differences between numpy and numexpr, except maybe `MultiplyNumberField` is slightly faster with numpy. So I'm happy using either for these cases.

lecoanet:
I just tested three different ways of doing spin recombination on our fiducial shell convection problem (Nmax=511, Lmax=110 on 1024 broadwell cores on Pleiades).

Python:
Total time: 225.8 sec
backward_spin_recombination + forward_spin_recombination = 1.89%
4.27 sec

Cython:
Total time: 226.1 sec
backward_spin_recombination + forward_spin_recombination = 1.22%
2.76 sec

Numexpr:
Total time: 227.7 sec
backward_spin_recombination + forward_spin_recombination = 1.64%
3.73 sec

In all cases, the spin recombinations are a pretty small cost. However, it looks like cython is decently faster. It's possible there is run-to-run variability, but in my experience, the variability is almost always in the transposes, not in the transforms.

Given this, I think it makes sense to stick with the cython implementation.

kburns:
Ok I think this might be ready to go.  I simplified things down to fast numpy operations for binary arithmetic, and cython only for the spin recombinations.  I then merged in the upstream d3 changes and patched up the spin recombinations assuming they're always out-of-place, since this is required for the cython methods.  This should now be easy to merge, and all d3 tests are passing.

lecoanet:
I noticed sometimes we use np.zeros, sometimes we use np.zeros_like. According to this, np.zeros should be faster:

https://stackoverflow.com/questions/27464039/why-the-performance-difference-between-numpy-zeros-and-numpy-zeros-like

Maybe we should switch everything to np.zeros? Don't know if it makes sense to do so as part of this PR or separately.

kburns:
I think that's pretty outdated.  On my laptop at least, the difference between `np.zeros` and `np.zeros_like` is pretty minor, like 10-20%.  A big improvement could come from changing these to `np.empty` or `np.empty_like` when we're just going to write in the data... but we probably want to think carefully about this and whether populating the "unused" portion with junk could ever hurt us.

lecoanet (approved):
Looks good and passes all tests.

PR #93, opened 2020-10-01
[d3] File handlers

jsoishi:
This will add evaluators to d3. It also adds support for virtual files, which allow on-the-fly creation of merged datasets 

### High Priority
- [x] fix bug with writing vectors reported by Alex
- [x] Produce virtual files only when a set is finished.
- [x] more stress-testing at high core counts and output cadence
- [x] Support for lower-dimensional outputs (slices and scalars)

### Low Priority
- [ ] Add old-style merging
- [ ] Maybe add option for virtual merge across sets (in time)
- [ ] add support for dropping empty files on cores that have no data (e.g. for scalars)

### DIFF
--- a/dedalus/core/distributor.py
+++ b/dedalus/core/distributor.py
@@ -230,7 +230,7 @@ class Layout:
             else:
                 # Block distribution otherwise
                 mesh = self.ext_mesh[axis]
-                if rank:
+                if rank is not None:
                     coords = np.zeros(self.dist.dim, dtype=int)
                     coords[~self.local] = self.dist.comm_cart.Get_coords(rank)
                     coord = coords[axis]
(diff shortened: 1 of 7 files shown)


kburns:
- [ ] Have an option or change the default to only produce virtual files when a set is finished.
- [ ] Add old-style merging
- [ ] Support for lower-dimensional outputs (slices and scalars)
- [ ] Maybe add option for virtual merge across sets (in time)

kburns:
I'd be happy to rebase and merge at this point -- it looks like Ben has been using this successfully for a while with ball simulations, and the cartesian tests for scalars and slices seem to be passing.  I think we should just get this in and then separately add old-style merging, merging in time, skipping empty files, and integral operators for the ball when we get around to them (hopefully soon).  How does that sound?

PR #94, opened 2020-10-09
add cartesian transpose

bpbrown:
Adds cartesian transpose operator, following spherical transpose operator implementation.  Adds implicit and explicit tests, and tests on both complex128 and float64 data types in 2-D and 3-D.  Moves spherical transpose operator from basis.py to operators.py and changes the logic of the dispatch.  Currently, this dispatches based on the coordinate system of the first index, and this could be generalized to multi-coordinate transposes in the future.


kburns:
Looks great!  Lets test squash & merge.

PR #96, opened 2020-10-16
Add Nonlinear Boundary Value Problems for d3

bpbrown:
Implements Nonlinear Boundary Value Problems (NLBVPs) in d3 and test of same in ChebyshevT and Legendre bases.  This differs slightly from the dedalus2 implementation, as we now differentiate the tau terms as well.  We solve `dH(u_n)du_{n+1} + dtau_{n+1}*P = -H(u_n) - tau_n*P`. The results should be the same, as `tau_{n+1} = tau_n + dtau_{n+1}` and this should lead to `dH(u_n)du_{n+1} + tau_{n+1}*P = -H(u_n)` as in dedalus2.

### DIFF
--- a/dedalus/core/arithmetic.py
+++ b/dedalus/core/arithmetic.py
@@ -467,7 +467,7 @@ class Product(Future):
 
     def matrix_dependence(self, *vars):
         coupling = self.matrix_coupling(*vars)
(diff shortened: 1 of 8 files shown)


bpbrown:
We revoked this pull request after realizing (post-merge) that it causes other tests to fail.

PR #97, opened 2020-10-16
Add Nonlinear Boundary Value Problems for d3

bpbrown:
Implements Nonlinear Boundary Value Problems (NLBVPs) in d3 and test of same in ChebyshevT and Legendre bases.  This differs slightly from the dedalus2 implementation, as we now differentiate the tau terms as well.  We solve `dH(u_n)du_{n+1} + dtau_{n+1}*P = -H(u_n) - tau_n*P`. The results should be the same, as `tau_{n+1} = tau_n + dtau_{n+1}` and this should lead to `dH(u_n)du_{n+1} + tau_{n+1}*P = -H(u_n)` as in dedalus2.

In the process of implementing NLBVP, we also implemented the Power operator.  We don't have a stand-alone Power operator test yet.

### DIFF
--- a/dedalus/core/arithmetic.py
+++ b/dedalus/core/arithmetic.py
@@ -467,7 +467,7 @@ class Product(Future):
 
     def matrix_dependence(self, *vars):
         coupling = self.matrix_coupling(*vars)
-        coupling[1] = True  # HACK HACK HACK for spheres coupling ell
+        #coupling[1] = True  # HACK HACK HACK for spheres coupling ell
         return coupling
 
     def matrix_coupling(self, *vars):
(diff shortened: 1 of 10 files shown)


bpbrown:
Fails `test_ball_ncc.py`

dedalus/core/arithmetic.py
```diff
@@ -467,7 +467,7 @@ def expression_matrices(self, subproblem, vars, **kw):
 
     def matrix_dependence(self, *vars):
         coupling = self.matrix_coupling(*vars)
-        coupling[1] = True  # HACK HACK HACK for spheres coupling ell
+        #coupling[1] = True  # HACK HACK HACK for spheres coupling ell
```
bpbrown, line 470: Uncommenting this line causes `test_ball_ncc.py` to work again and pass all tests, but this line does not work with `test_nlbvp.py`.

bpbrown:
All pytests in tests_new now pass.

bpbrown:
Rebased version passes all pytests.

PR #100, opened 2020-10-29
Disk basis and polar coordinates

jsoishi:
This implements a 2D disk basis and polar coordinates. It supports both real and complex fields and includes test coverage.


jsoishi:
All tests pass.

kburns (approved):
Looks good!

PR #104, opened 2020-11-20
[d3] EigenvalueProblems

bpbrown:
Implements Eigenvalue problems for d3 and 1d tests of same for Fourier and Chebyshev bases.

- [x] add sphere and disk test problems from papers
- [x] merge d3 in and check against test suite


kburns (approved):
Tests pass

jsoishi (approved):

PR #105, opened 2020-11-27
[d3] CFL operators

evanhanders:
- Basis objects now have the grid_spacing() method.
- CFL is computed using the grid-locked AdvectiveCFL operator, which returns a scalar frequency. Logic taking CFL from frequency -> timestep is the same as in d2.
- d3 calculates CFL on scales=1 grid regardless of dealiasing factor. This may change safety factor intuition from d2 sims where CFL uses scales=dealias grid.
- CFL tests on many bases added to tests_new/test_ivp.py.

### DIFF
--- a/dedalus/core/basis.py
+++ b/dedalus/core/basis.py
@@ -140,8 +140,16 @@ class Basis:
         shape[np.array(self.shape) == 1] = 1
         return tuple(shape)
 
-    def grid_spacing(self, **kwargs):
+    def global_grid_spacing(self, *args, **kwargs):
         """Global grids spacings."""
+        raise NotImplementedError
+
+    def local_grid_spacing(self, *args, **kwargs):
+        """Local grids spacings."""
+        raise NotImplementedError
+
+    def local_cfl_spacing(self, *args, **kwargs):
+        """Local CFL grid-crossing length scales."""
         raise NotImplementedError
 
     def global_grids(self, scales):
@@ -287,10 +295,24 @@ class IntervalBasis(Basis):
         super().__init__(coord)
 
     @CachedMethod
-    def grid_spacing(self, scale=None):
+    def global_grid_spacing(self, axis, scale=None):
         """Global grids spacings."""
         grid = self.global_grid(scale=scale)
-        return np.gradient(grid.flatten(), edge_order=2).reshape(grid.shape)
+        return np.gradient(grid, axis=axis, edge_order=2)
+
+    @CachedMethod
+    def local_grid_spacing(self, axis, scale=None):
+        """Local grids spacings."""
+        global_spacing = self.global_grid_spacing(axis, scale=scale)
+        if scale is None: scale = 1
+        local_elements = self.dist.grid_layout.local_elements(self.domain, scales=scale)[axis]
+        return reshape_vector(np.ravel(global_spacing)[local_elements], dim=self.dist.dim, axis=axis)
+
+    @CachedMethod
+    def local_cfl_spacing(self, axis, scale=None):
+        """Local CFL grid-crossing length scales."""
+        if scale == None: scale = 1
+        return self.local_grid_spacing(axis, scale) * scale
 
     # Why do we need this?
     def global_grids(self, scales=None):
@@ -521,20 +543,18 @@ class Jacobi(IntervalBasis, metaclass=CachedClass):
         matrix = convert @ matrix
         return matrix[:N, :N]
 
-
     @CachedMethod
-    def grid_spacing(self, scale=None):
-        """Global Jacobi grids spacings."""
+    def local_cfl_spacing(self, axis, scale=None):
+        """Local CFL grid-crossing length scales."""
+        cfl_spacing = super().local_cfl_spacing(axis, scale=scale)
         #Special case for ChebyshevT (a=b=-1/2)
         if self.a == -1/2 and self.b == -1/2:
             if scale is None: scale = 1
-            N = self.grid_shape((scale,))[0]
-            i = np.arange(N)
+            N = len(np.ravel(self.local_grids(scales=(scale,))[0]))
+            i = np.arange(N).reshape(cfl_spacing.shape)
             theta = np.pi * (i + 1/2) / N
-            native_spacing = np.sin(theta) * np.pi / N
-            return native_spacing * self.COV.stretch
-        else:
-            super(Jacobi, self).grid_spacing(scale=scale)
+            cfl_spacing[:] = scale * self.COV.stretch * np.sin(theta) * np.pi / N
+        return cfl_spacing
 
 def Legendre(*args, **kw):
     return Jacobi(*args, a=0, b=0, **kw)
@@ -782,12 +802,14 @@ class ComplexFourier(IntervalBasis):
         return (2 * np.pi / N) * np.arange(N)
 
     @CachedMethod
-    def grid_spacing(self, scale=None):
-        """Global Fourier grids spacings."""
+    def local_cfl_spacing(self, axis, scale=None):
+        """Local CFL grid-crossing length scales."""
+        cfl_spacing = super().local_cfl_spacing(axis, scale=scale)
         if scale is None: scale = 1
         N = self.grid_shape((scale,))[0]
         native_spacing = 2 * np.pi / N
-        return native_spacing * self.COV.stretch
+        cfl_spacing[:] = scale * native_spacing * self.COV.stretch
+        return cfl_spacing
 
     @CachedMethod
     def transform_plan(self, grid_size):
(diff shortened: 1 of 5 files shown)


bpbrown:
@evanhanders what's the status on this PR?

evanhanders:
@bpbrown -- this PR is broken as of https://github.com/DedalusProject/dedalus/commit/64494454f3567a68438b92641337722498ca09dc and the updates this weekend to Chebyshevs, etc. have caused further merge conflicts in basis.py. Fixing it is on my to-do list for this week.

evanhanders:
@bpbrown and @kburns , this PR is now aligned with modern d3 and ready for review.

I did some tests at Ra = 2e9 and Ek = 6e-6 (characteristic large-scale flow Re ~ 2e3 and Ro ~ 0.2), using SBDF2, to try to understand if the grid_spacing in the ball is correct. If it is, we'd expect safety ~ 0.4 to be the right choice for this timestepper. I tested safety = [0.35, 0.4, 0.5, 0.6, 0.8]. This specific run seems to be timestepping stably for safety <= 0.5 and failing (subtly) for safety >= 0.6. Take a look around the outer boundary of the axial vorticity (top left panel) of these equatorial cuts:

safety0.8
safety0.6
safety0.5
safety0.4
safety0.35

...I would've expected safety = 0.8 to just blow up, but it actually just has this pictured horrible ringing event at the outer boundary that translates happily along the boundary as it timesteps and it timesteps stably for hundreds of freefall times.

Perhaps this is a slightly unfair comparison. I used a different timestamp for safety = [0.35, 0.4, 0.5] than I did for 0.6 and 0.8, because the instabilities were obvious for the latter two and I hunted a snapshot that had an instability. For comparison, here's what those look like at the same timestamp as the lower safeties.

safety0.8_comp
safety0.6_comp

kburns:
Hi Evan, can you rebase everything onto the current d3 and remove the merge commits?  For the tests, I think we wanted to look at various localized flows to test whether the radial factors being used in the CFL calculation are correct, e.g. to find the true CFL limits near the center.

evanhanders:
@kburns I think there's going to be more changes, so I was planning on rebasing on current d3 right before this PR gets merged in. Would you prefer I do it now as well? (Happy to do so). I'll chat with you on slack further about the radial factors via slack :).

dedalus/core/basis.py
```diff
@@ -2681,6 +2738,12 @@ def global_grid_colatitude(self, scale):
         theta = self._native_colatitude_grid(scale)
         return reshape_vector(theta, dim=self.dist.dim, axis=self.axis+1)
 
+    def grid_spacing(self, axis, scales=None, r=None):
+        """Global grids spacings."""
+        if scales is None: scales = (1,1)
```
kburns, line 2743: `scales` is not used

dedalus/core/basis.py
```diff
@@ -4969,4 +5051,83 @@ def operate(self, out):
...
 
+class CartesianAdvectiveCFL(operators.AdvectiveCFL):
+
+    input_coord_type = CartesianCoordinates
+    input_basis_type = (ComplexFourier, RealFourier, Jacobi)
+
+    @CachedMethod
+    def grid_spacing(self, velocity):
+        coordsys = velocity.tensorsig[0]
+        grid_spacing = np.zeros_like(velocity.data)
```
kburns, line 5062: `grid_spacing` should be a list of broadcastable 1D vectors -- should not need to be the full shape of `velocity.data`

dedalus/core/basis.py
```diff
@@ -4969,4 +5051,83 @@ def operate(self, out):
...
+            global_spacing = basis.grid_spacing(scale=basis.dealias[0])
+            if isinstance(global_spacing, np.ndarray):
+                local_elements = self.dist.grid_layout.local_elements(self.domain, scales=basis.dealias[0])[i]
+                grid_spacing[i] = reshape_vector(np.ravel(global_spacing)[local_elements] * basis.dealias[0], dim=self.dist.dim, axis=i)
+            else:
+                grid_spacing[i] = global_spacing * basis.dealias[0]
+        return grid_spacing
+
+    def compute_cfl_frequency(self, velocity, out):
+
```
kburns, line 5074: Remove blank line

dedalus/core/basis.py
```diff
@@ -4969,4 +5051,83 @@ def operate(self, out):
...
+    input_coord_type = CartesianCoordinates
+    input_basis_type = (ComplexFourier, RealFourier, Jacobi)
+
+    @CachedMethod
+    def grid_spacing(self, velocity):
+        coordsys = velocity.tensorsig[0]
+        grid_spacing = np.zeros_like(velocity.data)
+        for i, c in enumerate(coordsys.coords):
+            basis = velocity.domain.get_basis(c)
+            global_spacing = basis.grid_spacing(scale=basis.dealias[0])
```
kburns, line 5065: If the bases all provide `global_grid_spacing`, maybe they should also provide `local_grid_spacing` instead of doing the slicing here?

dedalus/extras/flow_tools.py
```diff
@@ -205,23 +200,26 @@ def compute_dt(self):
...
-            raise ValueError("Wrong number of components for domain.")
-        for axis, component in enumerate(components):
-            self.add_velocity(component, axis)
+        self.frequencies.add_task(freq, layout='g', scales=freq.domain.dealias)
+
+    def add_velocity(self, velocity):
+        """
+        Add grid-crossing frequency from a velocity vector.
+        
+        Arguments
```
kburns, line 215: Should be `Parameters` in numpy docstring convention

kburns (changes requested):
Looks good! I think we could make the language/ideas around `grid_spacing` a little more consistent: for some bases/axes, the returned value is a true grid spacing that scales with the provided `scale`, while for others it is not.  Maybe the `grid_spacing` functions should always return true grid spacings, and the AdvectiveCFL operators can either use those or alternatives (e.g. `r/mmax`), and then the difference will be more explicitly clear?  Happy to make those changes myself after merging though, if that's easier.  Not sure what the best form for the tests is, but something that just tests that all forms of the operator can `operate` successfully seems good enough for now, so I think this is good.

dedalus/core/basis.py
```diff
@@ -140,6 +140,10 @@ def grid_shape(self, scales):
         shape[np.array(self.shape) == 1] = 1
         return tuple(shape)
 
+    def grid_spacing(self, **kwargs):
```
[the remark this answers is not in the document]
evanhanders: Done. Bases now have global_grid_spacing() and local_grid_spacing() methods (per comment below). Bases also have local_cfl_spacing() methods, which may be different from the actual physical grid spacings (e.g., grid_spacing = gradient(grid), cfl_spacing = r/m for the azimuthal part of the disk basis.

dedalus/core/basis.py
```diff
@@ -1774,6 +1817,20 @@ def global_grids(self, scales=None):
...
+        r = self.radial_COV.problem_coord(self._native_radius_grid(scale))
+        return reshape_vector(r, dim=self.dist.dim, axis=self.axis+1)
+
+    @CachedMethod
+    def grid_spacing(self, axis, scales=None):
+        """Global grids spacings."""
+        if scales is None: scales = (1,1)
+        if self.dims[axis] == 'azimuth':
+            # Scale grid spacing [ 1 / (mmax + 1) ] by the outer radius of the domain
+            return self.radius / (self.mmax + 1)
```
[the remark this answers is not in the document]
evanhanders: Done, see https://github.com/evanhanders/dedalus/blob/d3_cfl/dedalus/core/basis.py#L1868

dedalus/core/basis.py
evanhanders: scales is now passed through to local_grid_spacing (https://github.com/evanhanders/dedalus/blob/d3_cfl/dedalus/core/basis.py#L2800); this makes sure the returned array is the right shape for the input scales.

dedalus/core/basis.py
evanhanders: Done, I think. This un-vectorizes the division by spacing in compute_cfl_frequency() here: https://github.com/evanhanders/dedalus/blob/d3_cfl/dedalus/core/basis.py#L5147 (and makes it a for-loop).

dedalus/core/basis.py
```diff
@@ -4969,4 +5037,114 @@ def operate(self, out):
...
+    @CachedMethod
+    def cfl_spacing(self):
+        #Assumes velocity is a 3-length vector over spherical coordinates.
+        basis   = self.input_basis
+        dealias = basis.dealias
+        if isinstance(basis, BallBasis):
+            spacings = S2AdvectiveCFL.cfl_spacing(self, r=basis.radial_basis.radius)
+        elif isinstance(basis, ShellBasis):
+            spacings = S2AdvectiveCFL.cfl_spacing(self, r=1)
+            spacings[0] = spacings[0] * basis.local_grid_radius(dealias[2]) #get proper radial scaling for shell
```
kburns, line 5140: Why multiply separately instead of passing `local_grid_radiu` to `S2AdvectiveCFL.cfl_spacing`?

kburns:
I think we still need an annulus operator and test, but we can add that later if you don't have a chance now.

dedalus/core/basis.py
evanhanders: cfl_spacing is a CachedMethod, and I find that passing in a numpy ndarray as a kwarg throws this error:

```
    def __call__(self, *args, **kw):
        # First try direct call signature
        direct_call = (args, tuple(kw.items()))
        try:
>           return self.cache[direct_call]
E           TypeError: unhashable type: 'numpy.ndarray'
```

I want to make sure that cfl spacings are cached to avoid unnecessary computations. Thoughts on how to handle this?

kburns: Ah ok, yeah looks good then!

kburns (approved):

kburns:
Looks great, thanks a lot!

PR #107, opened 2020-12-03
Updated documentation for Compute Canada

mboisson:
The existing documentation for Cedar was outdated. I created one for Compute Canada as a whole. Building from source is not needed by our users, since we build wheels ourselves.


kburns:
Thanks!

PR #113, opened 2021-01-18
Fix typos in Tutorials

kuchtact:
Changed the comment explaining the operator object created in the tutorial. In v2.2006 the new operator added 10 and the field. In the new tutorial, the operator does something different.

### DIFF
--- a/docs/notebooks/KelvinHelmholtzInstability.ipynb
+++ b/docs/notebooks/KelvinHelmholtzInstability.ipynb
@@ -68,7 +68,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "#Aspect ratio 2\n",
+    "# Aspect ratio 2\n",
     "Lx, Ly = (2., 1.)\n",
     "nx, ny = (192, 96)\n",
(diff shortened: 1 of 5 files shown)


kburns:
Thanks for taking the time to find and fix these!  Happy to merge any time -- but don't want to do it too soon if you're still working through some notebooks and/or think you'll have other corrections to push.  Just let us know, and thanks again.

kuchtact:
I'm looking through the 'Example Notebooks' at the moment and was looking for input on how to format the notebook sections. Both notebooks have sections talking about 'Domains' and other similar things. These are better described in the 'Tutorial Notebooks' and trying to keep the explanations in all the notebooks the same quality would be a lot of copy pasting. Any suggestions as to what should be done?

kuchtact:
Everything should be ready to be merged now. I think the `TaylorCouetteFlow.ipynb` notebook still needs some fixing up but it's mostly stylistic and clarification rather than typos. That can be for another time and another pull request!

kburns:
Thanks!

Issue #109, opened 2021-01-08
[d3] Ball simulations fail with empty subsystems


jsoishi:
As is, with 256 cores specified on a `[16, 16]` grid, this script fails with error below.
If you use 64 cores with `[8,8]` it also fails, but works on 16 cores with `[4,4]`

```
Traceback (most recent call last):
  File "test_new_marti_dynamo.py", line 319, in <module>
    solver.step(dt)
  File "/home1/joishi1/vc/dedalus-dev/dedalus/core/solvers.py", line 387, in step
    self.timestepper.step(dt, wall_time)
  File "/home1/joishi1/vc/dedalus-dev/dedalus/core/timesteppers.py", line 140, in step
    evaluator.evaluate_scheduled(wall_time=wall_time, sim_time=sim_time, iteration=iteration)
  File "/home1/joishi1/vc/dedalus-dev/dedalus/core/evaluator.py", line 104, in evaluate_scheduled
    self.evaluate_handlers(scheduled_handlers, wall_time=wall_time, sim_time=sim_time, iteration=iteration, **kw)
  File "/home1/joishi1/vc/dedalus-dev/dedalus/core/evaluator.py", line 122, in evaluate_handlers
    self.require_coeff_space(fields)
  File "/home1/joishi1/vc/dedalus-dev/dedalus/core/evaluator.py", line 174, in require_coeff_space
    self.dist.paths[index-1].decrement(current_fields)
  File "/home1/joishi1/vc/dedalus-dev/dedalus/core/distributor.py", line 530, in decrement
    self.decrement_group(fields)
  File "/home1/joishi1/vc/dedalus-dev/dedalus/core/distributor.py", line 628, in decrement_group
    data0.append(field.data.reshape(flat_comp_shape))
ValueError: cannot reshape array of size 0 into shape (2,0,32)
```

jsoishi:
It appears that by changing `Lmax = 31` to `Lmax = 30` and 

```
b = basis.BallBasis(c, (2*(Lmax+1), Lmax+1, Nmax+1), radius=radius, dtype=dtype)
```

to 

```
b = basis.BallBasis(c, (2*(Lmax+2), Lmax+1, Nmax+1), radius=radius, dtype=dtype)
```

this code will work at [16,16] mesh. However, this is just hiding the problem which probably has to do with empty pencils in spheres.

bpbrown:
This is also a problem for test_new_marti_conv.py

lecoanet:
[...]

PR #116, opened 2021-02-04
[d3] Support empty subsystems in the ball

kburns:
This PR adds some quick fixes to get spherical problems working when we have empty subsystems, addressing #109.  I've done scaled down local tests, but haven't tested the original problem (test_new_mart_dynamo).


jsoishi (approved):
It works!

lecoanet:
I tried running test_new_marti_dynamo.py and it seemed to run & give the same answer on [8,8], [8,16], [16,8], and [16,16] processor meshes. I also tried running test_new_marti_conv.py, and it also seemed to work on up to 32 processors.

Issue #45, opened 2018-08-31
Future does not provide abstract meta_* methods


kburns:
**Original report ([archived issue](https://DedalusProject.github.io/dedalus_bitbucket_history/#!/dedalus-project/dedalus/issues/45)) by J. S. Oishi (Bitbucket: [jsoishi](https://bitbucket.org/%7Bab7c2db2-7642-47a8-9264-0939f601d904%7D/)).**

The original report had attachments: [mwe_attr_err.py](https://DedalusProject.github.io/dedalus_bitbucket_history/data/repositories/dedalus-project/dedalus/issues/45/attachments/mwe_attr_err.py)

----------------------------------------
When constructing objects like operators that inherit from Future, one needs to specify all of the possible metadata types as classes. As of right now, there are no abstract methods for these. As a result, when one creates a new Operator that is also FutureField, if you do not know to add `meta_parity` (for example), you will get a very obscure error, because Dedalus will attempt to resolve `meta_parity` in `UnaryGridFunctions`. 

This is very hard to track down, and would be greatly eased if `Future` provided abstract (but not implemented) methods for all possible metadata attributes on incoming fields. Then, at least one would know what to add.

kburns:
Fixed in https://github.com/DedalusProject/dedalus/commit/b4a77fa3fc9cebbf5fb18812d21ea971e15faf7e.

PR #131, opened 2021-03-13
Add abstract methods to FutureField, LinearOperator, and SpectralOperator

### LINKED ISSUE #130: [d3] Use abstract base classes to define interfaces in operators.py
I was checking in on #118, and noticed that in `operators.py` there's a [comment](https://github.com/DedalusProject/dedalus/blob/06dd8ac08a132fb48c0cd4f2202e32889646884c/dedalus/core/operators.py#L728-L745) that says that certain attributes have to be defined for subclasses of `SpectralOperator`, `LinearOperator`, and `FutureField`.  

A way to refactor this would be by using [abstract base classes](https://docs.python.org/3/library/abc.html).  The parent class would inherit from `abc.ABC` (short for **A**bstract **B**ase **C**lass), and the required attributes/methods would be decorated with `@abc.abstractmethod`.  Subclasses would have to implement each of the attributes...otherwise they would not be able to be instantiated.

```Python
>>> from abc import ABC, abstractmethod

>>> class AbstractInterface(ABC):
...     @abstractmethod
...     def required_attribute(self):
...         ...

>>> class FirstSubclass(AbstractInterface):  # does not define the required attribute
...     pass
>>> FirstSubclass()
TypeError: Can't instantiate abstract class FirstSubclass with abstract method required_attribute

>>> class SecondSubclass(AbstractInterface):
...     def required_attribute(self):  # subclass defines the required attribute
...         return 42

>>> SecondSubclass()
<__main__.SecondSubclass object at 0x7f77a5b4e670>
```
Defining the abstract interface in code is preferable to defining it in comments, as comments have a nasty habit of not getting updated when the corresponding code changes.  They're also a way of enforcing that interfaces get adhered to.  (Also, they end up being really fun to use once we get used to them.)

namurphy:
This PR is a model implementation of #130.  There were comments related to `FutureField`, `LinearOperator`, and `SpectralOperator` stating that certain attributes had to be defined by all subclasses.  This PR defines those attributes as abstract methods (after having `FutureField` inherit from `abc.ABC`.  Subclasses of these abstract base classes must override these abstract methods in order to be able to be instantiated.  This PR moves the interface from being described in comments to being implemented and enforced in the code itself.  

Closes #130.  Putting it in draft mode only because I haven't run the tests.

### DIFF
--- a/dedalus/core/future.py
+++ b/dedalus/core/future.py
@@ -3,6 +3,7 @@ Classes for future evaluation.
 
 """
 
+from abc import ABC, abstractmethod
 import numpy as np
 from functools import partial
 
@@ -249,12 +250,7 @@ class Future(Operand):
     #     return order
 
 
-
-
-
-
-
-class FutureField(Future):
+class FutureField(Future, ABC):
     """Class for deferred operations producing a Field."""
     future_type = Field
 
@@ -275,6 +271,18 @@ class FutureField(Future):
             return arg
         else:
             return FieldCopy(arg)
+
+    @abstractmethod
+    def domain(self):
+        pass
+
+    @abstractmethod
+    def tensorsig(self):
+        pass
+
+    @abstractmethod
+    def dtype(self):
+        pass
 
 
 class FutureLockedField(Future):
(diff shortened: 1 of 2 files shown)


namurphy:
Oh, and it looks like this might address #45 too.

PR #132, opened 2021-03-15
Left eigenvectors in solve_sparse

afraser3:
Following up on [this post in the google group from July](https://groups.google.com/g/dedalus-users/c/_10UjrWxEaI/m/gH5fTb1TAgAJ) (whoops), this adds an option to `EigenvalueSolver.solve_sparse()` for calculating left eigenvectors. Both `solve_sparse` and `solve_dense` include the "modified left eigenvectors", which are the dual basis to the right eigenvectors. Opening as a draft because I still want to double-check two minor things. I know Keaton has worked with these vectors before as well, so maybe he has a slicker implementation.

I used these heavily in a recent project, so if folks are interested I am happy to craft some sort of example script or notebook that uses these vectors.


afraser3:
By the way, as it's currently set up, you get that `np.matmul(solver.modified_left_eigenvectors.T, solver.eigenvectors)` is proportional to the Kronecker delta. Note there is NOT a conjugate on the first argument. To avoid confusion for people who are expecting orthogonality under your standard complex inner product, perhaps solver.modified_left_eigenvectors should have an overall np.conj() on it.

afraser3:
Attached is a simple demonstration, which does a sparse solve for a few modes for a KH-unstable flow in MHD (target=0 so it ends up finding continuum modes, not the unstable mode, in this case) and then demonstrates the biorthogonality of the sets of vectors.
[MHD_KH_EVP.py.zip](https://github.com/DedalusProject/dedalus/files/6142405/MHD_KH_EVP.py.zip)

dedalus/core/solvers.py
```diff
@@ -133,8 +139,23 @@ def solve_sparse(self, pencil, N, target, rebuild_coeffs=False, **kw):
         # Solve as sparse general eigenvalue problem
         A = pencil.L_exp
         B = -pencil.M_exp
+        # Solve for the right eigenvectors
         self.eigenvalues, self.eigenvectors = scipy_sparse_eigs(A=A, B=B, N=N, target=target, matsolver=self.matsolver, **kw)
+        if left == True:
+            # Solve for the left eigenvectors
+            #TODO: I think A and B are sparse.csr_matrix objects, in which case csr_matrix.getH might be better than np.conjugate(np.transpose()). Need to test.
```
kburns, line 146: Yes the `getH` method should be used here.

dedalus/core/solvers.py
```diff
@@ -133,8 +139,23 @@ def solve_sparse(self, pencil, N, target, rebuild_coeffs=False, **kw):
...
+            self.left_eigenvalues, self.left_eigenvectors = scipy_sparse_eigs(A=np.conjugate(np.transpose(A)),
+                                                                              B=np.conjugate(np.transpose(B)),
+                                                                              N=N, target=np.conjugate(target),
+                                                                              matsolver=self.matsolver, **kw)
+            # The following isn't an error, just bad luck: sometimes the sparse solver doesn't hit the same targets for left as it does for right eigenvectors
```
kburns, line 151: Maybe we should raise here?  Or only return the subset of modes that match?

kburns: Also is this mismatch ever just a sorting issue?  Maybe we need to compare after sorting.

dedalus/core/solvers.py
```diff
@@ -133,8 +139,23 @@ def solve_sparse(self, pencil, N, target, rebuild_coeffs=False, **kw):
...
+            if not np.allclose(self.eigenvalues, np.conjugate(self.left_eigenvalues)):
+                logger.warning("Conjugate of left eigenvalues does not match right eigenvalues.")
+            # In absence of above warning, modified_left_eigenvectors forms a biorthogonal set with the right eigenvectors
+            # (perhaps @ instead of * is better)
+            self.modified_left_eigenvectors = np.transpose(np.conjugate(self.left_eigenvectors.T)*-pencil.M)
```
kburns, line 156: Maybe we want a different sign and/or normalization here to make the modified left eigenvectors orthonormal the right eigenvectors?

kburns (changes requested):
Thanks! Here are a few comments on the code. Maybe we should chat to just discuss which conventions and normalizations might make the most sense, etc.

dedalus/core/solvers.py
afraser3: It's such a rare occurrence that I don't have the clearest recollection of the details, but from what I do remember, it's not just a sorting issue. I wish I had saved the example where this happened. I think raising is a reasonable solution except I can imagine at least one edge case where I wouldn't care about the error: suppose I want to use just one particular left eigenvector to calculate something, and I find that setting some particular target and Nev=3, I consistently find the correct left eigenvector as the third mode, whereas the first two modes are junk/mismatches that I don't care about. On the one hand, not raising risks misleading some users; on the other hand, raising means in that edge case I would have to keep fishing around to avoid raising the error.

kburns: It seems like a good option might be to raise on default, and add a keyword option to disable this and warn instead, so you can continue if you're aware of this issue and don't mind?

afraser3: Great idea! I'll work on implementing that.

afraser3: Now raises unless keyword `raise_on_mismatch = False`. Should it also sort the modes & eigenvalues before this point?

dedalus/core/solvers.py
afraser3: Latest commit demonstrates a few different options for what to do here (except the normalization part). Next time we meet we can chat about what makes the most sense.

afraser3:
@jsoishi, in defining the "modified left eigenvectors" (which form a biorthogonal set with the usual "right eigenvectors"), there's some freedom in how you define them, e.g., whether they're orthogonal to the right eigenvectors under a complex dot product or a non-complex dot product (see the block of comments in solve_sparse). Keaton and I were wondering if you had any thoughts here on whether there's any clearly-better conventions to follow. Also, whether you have any thoughts about things to put into set_state for the sake of playing well with eigentools.

afraser3:
As it currently stands, aside from some cleanup and docstring work that I need to do, I'm happy with what this code does.

dedalus/core/solvers.py
```diff
@@ -133,22 +152,94 @@ def solve_sparse(self, pencil, N, target, rebuild_coeffs=False, **kw):
...
+            self.left_eigenvalues, self.left_eigenvectors = scipy_sparse_eigs(A=A.getH(),
+                                                                              B=B.getH(),
+                                                                              N=N, target=np.conjugate(target),
+                                                                              matsolver=self.matsolver, **kw)
+            # TODO: make it so that left_eigenvectors is Y^* (do the same for eigenvalues)?
+            # TODO: and modified_left_eigenvectors is Y^*M?
+            # TODO: Maybe start with making it consistent with scipy.linalg.eig formalism, with clear docstring, then ask Jeff Oishi for input
+            # TODO: after normalizing, one last iteration to clean up orthogonality? -- kick this can down the road, Keaton will look into it
+            # TODO: decide whether and how to sort eigenvalues and eigenvectors before this point.
+            if not np.allclose(self.eigenvalues, np.conjugate(self.left_eigenvalues)):
```
afraser3, line 172: Hey Keaton, Zach Williams (a collaborator and the current number one user of this functionality I suspect) is telling me that, even with sorting, these mismatches are a near constant problem for him. If he passes the `target` kwarg, it's not as bad for the modes near the target, but he's often searching for 10-100 modes at once, and it sounds like the mismatch is a significant obstacle for him.

Any idea what can be done? Is this an inherent consequence of using an iterative solver? Would tinkering with any of the matrix solvers in dedalus.cfg affect this?

jsoishi: Is the mismatch occuring because the order is wrong or because the left eigenvalues/vectors are inaccurate? If it's the latter, is the problem just that he doesn't have resolution on those modes?

afraser3: Hmm good question. He's looking into it now. Part of it is certainly the order, but it sounds like he has no problem sorting the modes himself. He finds that, even after sorting, there are eigenvalues in one set that don't correspond to eigenvalues in the other set.

afraser3: Seems like a good argument for me figuring out how to put this into eigentools!

jsoishi:
> @jsoishi, in defining the "modified left eigenvectors" (which form a biorthogonal set with the usual "right eigenvectors"), there's some freedom in how you define them, e.g., whether they're orthogonal to the right eigenvectors under a complex dot product or a non-complex dot product (see the block of comments in solve_sparse). Keaton and I were wondering if you had any thoughts here on whether there's any clearly-better conventions to follow. Also, whether you have any thoughts about things to put into set_state for the sake of playing well with eigentools.

@afraser3 Re the choice:This is tricky. I kind of think the orthogonality should be under the standard (i.e. complex) inner product.

dedalus/core/solvers.py
afraser3: Ok, Zach has looked into this. There was an issue where a mistake in the sorting caused apparent mismatches that were eventually cleared up. Even after fixing that, there were still some mismatches, but now we know why. He's looking at EVPs corresponding to the tearing instability where the eigenvalue spectrum is symmetric along one axis in the complex plane (see figure 1 [here](https://arxiv.org/abs/1808.10208) and note the symmetry with respect to reflections over the horizontal axis). So it seems that if your target/guess eigenvalue passed to the solver is equidistant to two different eigenvalues due to this symmetry, the solver sometimes finds one frequency for the right modes and the other frequency for the left modes. At least, that's our working hypothesis for what's going on.

kburns:
I think we've all let this languish a bit, but can we remind ourselves where it currently stands?  I'd like to put out another release this month with some longstanding fixes and performance improvements, and it would be nice to include this work if we're ready to wrap it up with a little more work.

afraser3:
I think this is now ready to roll! There are three minor items I specifically wanted to point out to you @kburns 

[This comment](https://github.com/afraser3/dedalus/blob/ec91650e7affd46984e623ffe593d3a6e1b16beb/dedalus/core/solvers.py#L173) stems from an idea you proposed a while back. I'm not really sure how to go about doing this, but I wanted to point it out to you again in case you want to tackle it.

[This comment](https://github.com/afraser3/dedalus/blob/ec91650e7affd46984e623ffe593d3a6e1b16beb/dedalus/core/solvers.py#L188) is pointing out a use of * that is maybe sloppy coding because it's a bit ambiguous what's happening, and I'm wondering if there's an obvious better replacement. But it does the job.

[This comment](https://github.com/afraser3/dedalus/blob/ec91650e7affd46984e623ffe593d3a6e1b16beb/examples/evp/1d_MHD_KH_left_eigenvectors/MHD_KH_EVP.py#L58) stems from me not understanding what rebuild_coeffs does -- should I leave it as is or change it in the subsequent line?

Otherwise I think it should be all set! Thanks for letting me contribute this, even though I know you also had similar code of your own that did the same thing. It was fun to work on a PR and get a feel for how to contribute to big open source projects.

examples/evp/1d_MHD_KH_left_eigenvectors/MHD_KH_EVP.py
```diff
@@ -0,0 +1,103 @@
+"""
+Runs an eigenvalue problem for the KH instability of a U = tanh(z) shear layer in MHD
+for purposes of testing left eigenvectors
+
+TODO: Remove this script before merging with main. This is just meant for testing.
```
kburns, line 5: Do you want to keep this in? If so, I think it needs a little cleanup (particularly the comments). Otherwise it could also be reimplemented as a test.

examples/evp/1d_MHD_KH_left_eigenvectors/MHD_KH_EVP.py
```diff
@@ -0,0 +1,103 @@
...
+problem.add_equation("phi_z - dz(phi) = 0")
+problem.add_equation("psi_z - dz(psi) = 0")
+
+problem.add_bc("left(phi) = 0")
+problem.add_bc("right(phi_z) = 0")
+problem.add_bc("right(Bz) = 0")
+solver = problem.build_solver()
+
+if solve_sparse:
+    # Keaton: I don't know what rebuild_coeffs actually does -- should it be True or False here? Does it matter?
```
kburns, line 58: This rebuilds the matrices if you've changed the parameter values -- shouldn't be necessary here.

kburns (commented):
Would be great to finally get this merged! Here's just a couple quick points.

examples/evp/1d_MHD_KH_left_eigenvectors/MHD_KH_EVP.py
afraser3: I think it's helpful as a demonstration of the biorthogonality relations, but seems like too minor/redundant of a thing to leave as an example script. I like the idea of implementing it as a test, but I'm not sure what that entails or how best to do that. Maybe we can chat further about this via email or zoom.

jsoishi: @afraser3 we can chat about this on Thursday if you'd like...

afraser3: Evan walked me through a little bit of how to implement tests. I'll give it a shot tomorrow and see what happens. If it doesn't work out, I think it's reasonable to cut this from the PR.

afraser3: Properly testing bi-orthogonality is just a little more finicky than I realized, I think because of sorting left vs right eigenmodes and the inf/nans you get when doing dense solves. Is it fine to just cut the example altogether, or is that bad practice?

kburns: I think it's ok, we can always pull this example out later if needed. Thanks!

PR #141, opened 2021-05-17
[d3] Annulus

kburns:
This PR includes the annulus class and a lot of restructuring of disks.


kburns:
A few operators still need to be implemented for the polar bases (e.g. component selection), but I think we should just merge this in for now.

PR #143, opened 2021-05-23
Update tutorial notebooks

RcrdPhysics:
solving warnings and update functions applied in some examples


kburns:
Hello, it looks like some wires got crossed here on the git side, since the PR contains what seems to be a copy of all the source code files out of the `dedalus` folder and directly into the repo.  This also makes it difficult to see/assess any proposed changes at this stage.  I'm going to add the "invalid" flag for now to ensure this doesn't accidentally get merged before the file arrangement is fixed up, but I'll be happy to review the changes after that!

PR #144, opened 2021-06-09
[d3] Exclude non-existent LHS components for low ell

kburns:
This PR modifies the LHS construction to drop the blocks associated with tensor components that do not exist at low ell.  This should allow us to rewrite most of the example equations without as many conditions on ell ~ 0.

### DIFF
--- a/dedalus/core/basis.py
+++ b/dedalus/core/basis.py
@@ -2782,9 +2782,12 @@ class SpinWeightedSphericalHarmonics(SpinBasis):
         enum_components_output = []
         for i, comp in enum_components_input:
             # Filter for indices in self.coordsystem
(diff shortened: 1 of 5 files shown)


kburns:
@lecoanet where did we leave off here?  I think we still need to handle 2d vs 3d surface vector fields for BCs to behave properly, at least.

lecoanet:
If I remember correctly, there were issues with selecting the correct components from the ell=0 radial or angular BC's. I think it selects the correct components for an ell=0 3D BC. In particular, we would like to remove the condition on this BC:
`problem.add_equation(eq_eval("u_perp_bc = 0"), condition="ntheta != 0")`
In the marti conv test script.

kburns:
Should be fixed with c3dd7c8.

PR #147, opened 2021-06-16
Update dedalus_tutorial_fields_operators.ipynb

deKeijzer:
Fix for `AttributeError: module 'dedalus.public' has no attribute 'Field'`


kburns:
This error only comes up if you try to run the latest examples (from master) using an older version of the code (e.g. the latest pip release).  Everything should execute correctly if you use the same version of the code and examples, so this reversion shouldn't be necessary.

PR #149, opened 2021-07-26
[d3] Angular interpolation

kburns:
This PR implements operators for interpolation in azimuth and colatitude for spherical and polar bases.  These are implemented as collocation operators, and lock the resulting fields to grid space in the corresponding axes to prevent regularity violations in the forward transforms.  The resulting fields cannot be transformed, differentiated, or have their scales changed, but can be used for output on the dealias scales.


dedalus/core/distributor.py
```diff
@@ -395,6 +396,7 @@ def decrement_single(self, field):
         cdata = field.data
         # Transform non-constant bases with local data
         if (basis is not None) and np.prod(gdata.shape):
+            (gdata.shape, cdata.shape, field.domain.bases[0].shape, self.layout0.grid_space)
```
lecoanet, line 399: What is this doing?

kburns: Oops looks like its just a leftover `print` fragment.

dedalus/tests/test_spherical3D_operators.py
```diff
@@ -225,20 +226,111 @@ def test_interpolate_scalar(Nphi, Ntheta, Nr, k, dealias, dtype, basis, radius):
 @pytest.mark.parametrize('dealias', dealias_range)
 @pytest.mark.parametrize('basis', [build_ball, build_shell])
 @pytest.mark.parametrize('dtype', [np.float64, np.complex128])
-@pytest.mark.parametrize('radius', [0.5, 1.0, 1.5])
-def test_interpolate_vector(Nphi, Ntheta, Nr, k, dealias, dtype, basis, radius):
+@pytest.mark.parametrize('theta_interp', [0, np.pi/4, np.pi/2, np.pi])
```
lecoanet, line 229: What happens when we interpolate theta to 0 or pi?

dedalus/tests/test_spherical3D_operators.py
```diff
@@ -225,20 +226,111 @@ def test_interpolate_scalar(Nphi, Ntheta, Nr, k, dealias, dtype, basis, radius):
...
+
+
+@pytest.mark.parametrize('Nphi', [16])
+@pytest.mark.parametrize('Ntheta', [8])
+@pytest.mark.parametrize('Nr', Nr_range)
+@pytest.mark.parametrize('k', k_range)
+@pytest.mark.parametrize('dealias', dealias_range)
+@pytest.mark.parametrize('basis', [build_ball, build_shell])
+@pytest.mark.parametrize('dtype', [np.float64, np.complex128])
+@pytest.mark.parametrize('r_interp', [0.5, 1.0, 1.5])
```
lecoanet, line 248: Similarly, what would happen if we interpolate r=0?

dedalus/tests/test_spherical3D_operators.py
kburns: Right now the output remains a function of phi and r, and will just give identical values for all phi (but structurally still include all m's).

lecoanet:
Right now some transforms of interpolated data are supposed to fail, and others are supposed to be possible. It would be good to include some tests of this, i.e., forward/backward transform test for some interpolated data.

kburns:
I think the only allowed transform is the azimuth transform after colatitude interpolation, and the test covers this (backwards at least) since it compares the solution in grid space.

dedalus/core/distributor.py
kburns: Fixed

lecoanet (approved):
Looks good now!

PR #150, opened 2021-07-28
[d3] Integral operators

kburns:
This PR includes spectral integral operators and tests for the spherical and polar bases.


kburns:
Volume integration, spherical averaging, and azimuthal averaging all implemented and tested (explicitly).

lecoanet (approved):
Everything looks good. I also tested using the integration routine for the marti hydro test, and it works.

PR #159, opened 2021-10-07
[d3] Small optimizations for small problem sizes

lecoanet:
I replaced np.prod with prod is several parts of the code. I also allowed users to specify meshes with size 1 in certain directions, e.g. mesh=[1, 32]. This is useful when doing very anisotropic problems.

### DIFF
--- a/dedalus/core/distributor.py
+++ b/dedalus/core/distributor.py
@@ -87,6 +87,9 @@ class Distributor:
             comm = MPI.COMM_WORLD
         if mesh is None:
             mesh = np.array([comm.size], dtype=int)
+        else:
+            if isinstance(mesh, list) or isinstance(mesh, tuple):
+                mesh = np.array(mesh, dtype=int)
         # Trim trailing ones
         mesh = 1 + np.trim_zeros(mesh - 1, trim='b')
         self.dim = dim = len(self.coords)


bpbrown:
When testing this, I ran into an error anytime I passed a specified mesh of either list or tuple type:

```
Traceback (most recent call last):
  File "hydro.py", line 114, in <module>
    d = de.Distributor((c,), mesh=mesh, dtype=np.float64)
  File "/home5/bpbrown/conda_install/src/dedalus-d3/dedalus/core/distributor.py", line 92, in __init__
    mesh = 1 + np.trim_zeros(mesh - 1, trim='b')
TypeError: unsupported operand type(s) for -: 'list' and 'int'
```
or
```
Traceback (most recent call last):
  File "hydro.py", line 114, in <module>
    d = de.Distributor((c,), mesh=mesh, dtype=np.float64)
  File "/home5/bpbrown/conda_install/src/dedalus-d3/dedalus/core/distributor.py", line 92, in __init__
    mesh = 1 + np.trim_zeros(mesh - 1, trim='b')
TypeError: unsupported operand type(s) for -: 'tuple' and 'int'
```

This is fixed by commit 8d4d9cb47ca5b9accda10a2d04900aea2b6a896d by doing an instance check for list or tuple types, and then doing a `np.array()` on mesh.

Fails a bunch of ncc tests and transform tests, as @kburns had noted.

PR #161, opened 2021-10-10
[d3] New group specification strategy

kburns:
This PR implements a new system for finding the local groups and local group slices that works across multiple layouts. Bases now specify an `elements_to_groups` method that produces the group numbers (mode/quantum numbers) as a function of the global element indices for an arbitrary layout. The local element indices are determined entirely by the distributor, so this separates the data decomposition (the distributor's job) from the mode ordering/reordering (the basis's job). This fixes a number of previous issues with the parallel evaluation of operators in layouts other than full coeff space, and simplifies much of the basis code. Tests are passing and examples are running, besides NCC and some transform tests (failing due to unrelated aliasing issues).

### DIFF
--- a/dedalus/core/basis.py
+++ b/dedalus/core/basis.py
@@ -2093,7 +2093,7 @@ class DiskBasis(PolarBasis, metaclass=CachedClass):
 
     @staticmethod
     def _nmin(m):
-        return np.maximum(m//2, 0)
+        return abs(m) // 2
 
     def __add__(self, other):
         if other is None:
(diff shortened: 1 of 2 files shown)


bpbrown:
On the boussinesq mhd ball test script that I've been using,  commit 196ae1793240af2465e501a1159f03a63ea411b6 fixes the vectorization problem we were seeing in parallel.

kburns:
Ok just pushed some commits that should fix the startup slowdowns.

kburns:
@bpbrown Thanks! Just to confirm, can we double-check that the MHD problem still runs with the latest changes? They modified the earlier fixes for the "empty arrays in np.vectorize" problem -- so just want to make sure that didn't re-break.

kburns:
Great, thanks!

PR #162, opened 2021-10-19
[d3] Sphere basis

kburns:
This PR implements S2 operators, tests, and examples. The tests all currently pass except for a couple related to the sine mode for m=0, which is being addressed in the d3-valid branch. Remaining things:

- [x] Add cosine multiplication operator
- [x] Add more tests
- [x] Finalize examples


kburns:
Probably only need one of barotropic/shallow water examples, but need to understand nonlinear stability, and probably add diffusion, etc.

kburns:
Seems like the instability might just be from using complex numbers. Zeroing the imaginary part each timestep fixes the issue, and it runs as expected (maintains nonlinear solution with slow viscous decay).

kburns:
Decided to keep the shallow water example and add a jet to it, and move the barotropic script to the scratch directory, since its really a scalar problem that doesn't require vector fields.

I think this is pretty much ready to go now.  Could always use more tests, of course, but maybe good enough for now?

bpbrown:
Looks good to me; we can add other tests and examples later, so I'd say
let's get this merged!

PR #165, opened 2021-11-01
Hdf5 1.12

mboisson:
Dedalus does not support HDF5 1.10 (which is our default), it requires 1.12.


mboisson:
Dedalus does not support HDF5 1.10 (which is our default), it requires 1.12.

kburns:
Thanks!

PR #167, opened 2021-11-12
[d3] Built-in performance tracking

kburns:
This PR currently just adds some basic built-in timings for IVPs to reduce boilerplate and provide more consistent performance evaluations across scripts. The IVP solvers now internally time:

1) "Setup time" meaning everything from solver initialization to the beginning of the first iteration.
2) "Warmup time" meaning the time for the first N iterations, where N is adjustable and can be set at/after solver initialization, and defaults to 10.
3) "Run time" meaning the time for all iterations after N.

The `solver.log_stats` method can be used to easily log these at the end of an IVP, along with the "CPU time" of the simulation in cpu-hrs, and the "Speed" of the simulation in mode-stages/cpu-sec. Here "modes" is the total number of coeff-space degrees of freedom, calculated from adding the sizes of all the subproblem matrices (so the true DOF count even for spheres, etc.).  And "stages" is the total number of RHS evaluations, computed as the total iterations times the stages-per-iteration for the chosen timestepper. So this should put the multistep and RK methods closer to the same footing. The string formatting for these is customizable via a keyword argument to the method.

Any thoughts? This seems like a good core set of basic stats to report at the end of most simulations. Do we prefer the examples to use this method and have reduced boilerplate, or to be more explicit as before?

### DIFF
--- a/dedalus/core/solvers.py
+++ b/dedalus/core/solvers.py
@@ -402,15 +402,18 @@ class InitialValueSolver(SolverBase):
     # Default to factorizer to speed up repeated solves
     matsolver_default = 'MATRIX_FACTORIZER'
 
-    def __init__(self, problem, timestepper, enforce_real_cadence=100, **kw):
+    def __init__(self, problem, timestepper, enforce_real_cadence=100, warmup_iterations=10, **kw):
         logger.debug('Beginning IVP instantiation')
         super().__init__(problem, **kw)
         self.enforce_real_cadence = enforce_real_cadence
         self._wall_time_array = np.zeros(1, dtype=float)
-        self.start_time = self.get_wall_time()
+        self.init_time = self.get_wall_time()
         # Build subproblems and subproblem matrices
         self.subsystems = subsystems.build_subsystems(self)
         self.subproblems = subsystems.build_subproblems(self, self.subsystems, ['M', 'L'])
+        # Compute total modes
+        local_modes = sum(ss.subproblem.pre_right.shape[1] for ss in self.subsystems)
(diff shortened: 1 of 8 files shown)


kburns:
Can also discuss other performance tracking things that we've been thinking about (e.g. transform/transpose counts) here, or in another PR/issue.

jsoishi (commented):
This looks great!

bpbrown (approved):
Tested this extensively using the 2-d rayleigh benard scripts in the `rb_d3` repo and ball boussinesq hydro and mhd in the `mars` repo.  In all cases, I tested the internal perfomrnace tracking (hereafter "speed") against our previous standard DOF-cycles/cpu-sec outputs (hereafter "DOF").  In all cases, outputs from the new built in performance tracking agreed with those adhoc DOF estimates.

Some specific notes:

- All tests run on NASA/Pleiades using from 1--512 skylake cores.

- In `rb_d3` repo: in `rayleigh_benard_2d.py` I tested a variety of timesteppers.  Different multistep timesteppers had similar performances, and different multistage timesteppers had same performance metrics using the internal speed timer (and differed using the adhoc DOF calculation).

- in `mars` repo: testing `hydro.py` and `mhd.py`, both scripts report very similar speed values, but very different adhoc DOF values.  So speed appears to be correctly accounting for the difference in number of solved for variables between those two scripts.  DOF values were lower by same factor of solved for variables.


This looks good to me, and I think it's ready to merge.  Haven't seen an edge case or any problems.

kburns:
Thanks, Ben! I think the easiest thing will be to rebase this onto the valid_modes stuff after that's merged.

PR #168, opened 2021-11-24
[d3] Dropping invalid modes

kburns:
This PR implements a new system for determining which modes should be dropped from the LHS. This used to be done using `valid_components`, but this did not cover all necessary cases (like real variables in ell-coupled problems on S2) where the validity change the mode limits for certain subproblems and components, rather than eliminating them entirely. The new system adds `valid_elements` methods to the bases, which determine for every element and component, the array of valid elements. This is then used to arbitrarily eliminate modes from the LHS with projections in the preconditioners.

This system also allows us to drop the sine-like part of the m=0 modes in curvilinear coordinates. Currently this is only implemented for scalar fields, since this is necessary for allowing problems that couple true scalars (numbers, like gauges) into the problem. In the future, this could be further enhanced to drop the sin-like part of any full-0 spin component, and perhaps to enforce whatever symmetry that maps to in regularity space. This could eventually completely eliminate the need to perform transform loops to enforce Hermitian conjugacy.

For now, it looks like the tests are passing (excluding unrelated failures in NCC and transform tests) and the examples are all running. One nice feature is that this enables us to eliminate the need for equation conditions in most cases, replacing them with integral gauges, etc. This has been done for all example scripts.

### DIFF
--- a/dedalus/core/arithmetic.py
+++ b/dedalus/core/arithmetic.py
@@ -390,8 +390,17 @@ class Product(Future):
         ncc_data = self._ncc_data
         separability = ~ subproblem.problem.matrix_coupling
         #return = self._ncc_matrix_recursion(ncc_data, ncc.tensorsig, ncc.bases, operand.bases, operand.tensorsig, separability, self.gamma_args, **kw)
-        ncc_basis = ncc.domain.full_bases[-1]
-        arg_basis = operand.domain.full_bases[-1]
+        if ncc.domain.bases:
+            ncc_axis = ncc.domain.bases[-1].last_axis
+        else:
+            ncc_axis = 0
+        if operand.domain.bases:
+            arg_axis = operand.domain.bases[-1].last_axis
+        else:
+            arg_axis = 0
+        axis = max(ncc_axis, arg_axis)
+        ncc_basis = ncc.domain.get_basis(axis)
+        arg_basis = operand.domain.get_basis(axis)
         ncc_ts = ncc.tensorsig
         coeffs = ncc_data
         arg_ts = operand.tensorsig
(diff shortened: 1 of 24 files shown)


bpbrown (approved):
I've tested a variety of scripts in the new "valid-modes" formation.  These scripts are in the `rb_d3` repo, the `dstar` repo and the `mars` repo.  All scripts work with `d3-valid`.

Some notes on tests:
- `rb_d3/rayleigh_benard_2d_drag.py`: 2-d rayleigh benard script; works and tau errors are in line with previous experience.  The gauge enforcing `taup` is particularly small in resolved sims compared to other taus in the system.
- `mars/hydro.py`: ball boussinesq convection with internal heating.  Using `--benchmark`, `SBDF4` and `Ntheta=32`, `Nr=32`, and `dt=0.1` (equivalent to `dt=3e-5` in Lecoanet et al 2019) reproduces Marti benchmark with `E0=29.1204548887` which matches Lecoanet et al 2019 to all reported digits.
- `mars/mhd.py`: ball MHD convectively driven dynamo.  Runs; required modifications to potential field boundary conditions.  See `dedalus/tests/test_evp.py` for correct current form of potential field conditions.
- `dstar/mdwarf_hydro.py`: ball anelastic convection.  Runs.  Tests using gauge enforcing `τp` in a NCC setting.

These new "valid-modes" approaches in `d3-valid` really clean these scripts up, help with the equation representation, and give a handle on the divergence enforcement.  Performance on all modified scripts looks reasonable, though I have yet to do detailed performance/scaling testing on NASA/Pleiades.

kburns:
Thanks Ben!

PR #174, opened 2022-02-03
Fix CFL in spherically symmetric runs

evanhanders:
Fixes an edge case CFL bug when the user assumes azimuthal symmetry (1 azimuthal point).

### DIFF
--- a/dedalus/core/basis.py
+++ b/dedalus/core/basis.py
@@ -3940,7 +3940,11 @@ class Spherical3DBasis(MultidimensionalBasis):
     def global_grid_spacing(self, axis, scales=None):
         """Global grids spacings."""
         if scales is None: scales = (1,1,1)
-        return np.gradient(self.global_grids(scales=scales)[axis], axis=axis, edge_order=2)
+        grid = self.global_grids(scales=scales)[axis]
+        if grid.size == 1:
+            return np.array([np.inf,], dtype=grid.dtype)
+        else:
+            return np.gradient(grid, axis=axis, edge_order=2)
 
     @CachedMethod
     def local_grid_spacing(self, axis, scales=None):


dedalus/core/basis.py
```diff
@@ -3868,7 +3868,11 @@ def local_grids(self, scales=None):
     def global_grid_spacing(self, axis, scales=None):
         """Global grids spacings."""
         if scales is None: scales = (1,1,1)
-        return np.gradient(self.global_grids(scales=scales)[axis], axis=axis, edge_order=2)
+        grid = self.global_grids(scales=scales)[axis]
+        if len(grid.flatten()) == 1:
```
kburns, line 3872: "ravel" would be better than "flatten" here to avoid a copy

kburns: This could probably also just be `grid.size == 1`?

dedalus/core/basis.py
```diff
@@ -5458,7 +5462,9 @@ def cfl_spacing(self, r=None):
         if basis.Lmax == 0:
             s2_spacing[:] = np.inf
         else:
-            s2_spacing[:] = r / np.sqrt(basis.Lmax*(basis.Lmax + 1))
+            for i in range(len(s2_spacing.shape)):
```
kburns, line 5465: Would s2_spacing ever not have dim=2?

evanhanders:
@kburns I just updated this per your suggestions and tried it out in on the ball and shell example problems (setting nphi = 1 in each). Seems to work!

dedalus/core/basis.py
```diff
@@ -5534,9 +5534,7 @@ def cfl_spacing(self, r=None):
         if basis.Lmax == 0:
             s2_spacing[:] = np.inf
         else:
-            for i in range(len(s2_spacing.shape)):
-                if s2_spacing.shape[i] > 1:
-                    s2_spacing[i, :] = r / np.sqrt(basis.Lmax*(basis.Lmax + 1))
+            s2_spacing[:] = r / np.sqrt(basis.Lmax*(basis.Lmax + 1))
```
kburns, line 5537: I think we still want the check on the spacing shapes, right? To catch the case where Nphi=1 (so mmax=0 but Lmax > 0)?

dedalus/core/operators.py
```diff
@@ -1095,9 +1095,6 @@ def _preprocess_args(cls, operand, coord=None):
         # Integrate over all operand bases by default
         if coord is None:
             coord = [basis.coordsystem for basis in operand.domain.bases]
-        # Split Cartesian coordinates
-        if isinstance(coord, coords.CartesianCoordinates):
-            coord = coord.coords
```
kburns: Was this meant to be removed?

dedalus/core/operators.py
```diff
@@ -1165,9 +1162,6 @@ def _preprocess_args(cls, operand, coord=None):
         # Average over all operand bases by default
         if coord is None:
             coord = [basis.coordsystem for basis in operand.domain.bases]
-        # Split Cartesian coordinates
-        if isinstance(coord, coords.CartesianCoordinates):
-            coord = coord.coords
```
kburns: Was this meant to be removed?

kburns (changes requested):
I think something funny happened with rebasing or merging -- looks like this PR contains and undoes the latest commit on master?

evanhanders:
Hi @kburns; yes something went weird with that most recent commit and rebasing. PR should now be properly rebased onto dedalus/master. Will take a look at your other comment shortly and test on an azimuthally symmetric run.

dedalus/core/basis.py
evanhanders: I don't think we do. But this may reflect a different design choice that should be made here.

We still want to preserve the r / sqrt(L*(L+1)) behavior of the CFL in the theta direction, and this current formulation does that in both BallBases and ShellBases. If s2_spacing were setting something that depended on phi and theta, then we'd want to be more careful, but since it's just setting a scalar value I think the only conditioning necessary is when Lmax = 0.

kburns: Ah, right, because we're using the magnitude of the horizontal velocity to compute the CFL, not the individual components. That will just mean that there's a CFL restriction associated with zonal velocities in axisymmetric problems. Maybe not a big deal for now, but maybe that's something we eventually want to fix?

evanhanders: Ah! Good point, I see what you're saying and we should probably address it here. I think this calls for a minor redesign (s2_spacing should have a (ntheta, nphi) shape)). I'll think about that a bit and update it within a couple hours, hopefully.

kburns: We have to think about it a bit carefully. Say you have an axisymmetric (Nphi=1) velocity field V that you use to advect a non-axisymmetric (Nphi > 1) scalar T. In that case there will still be an azimuthal CFL limit, but we can't infer the right grid spacing to use for it from the bases of the velocity field itself...

evanhanders: Hmm. Ok, on second thought, this is feeling like it's beyond the scope of this PR (which is to make it so ntheta = 1 balls/shells don't crash on initialization), and that's more of a new feature? Maybe we should open a separate issue?

evanhanders: Yep! Merge away, CFL seems to be initializing in both parallel and serial, for balls and shells.

PR #175, opened 2022-02-15
Added plot showing sorted eigenvectors

hdrake:
I added a few lines so that users can see how to plot (in grid space) the eigenvectors with the lowest/highest growth rates.

It took me a while to figure out the new `set_state` method API that requires specifying the `subsystem` in addition to the index, so I thought it would be useful to include it in at least one of the EVP examples.


examples/evp_1d_waves_on_a_string/waves_on_a_string.py
```diff
@@ -67,3 +67,11 @@
 plt.tight_layout()
 plt.savefig('eigenvalue_error.pdf')
 
+plt.figure(figsize=(6, 4))
+for n, idx in enumerate(np.argsort(solver.eigenvalues)[:5], start=1):
```
kburns, line 71: `x` is not defined yet. Could add `x = dist.local_grid(xbasis)` before the for loop.

examples/evp_1d_waves_on_a_string/waves_on_a_string.py
```diff
@@ -67,3 +67,11 @@
...
+    solver.set_state(idx, solver.subsystems[0])
+    plt.plot(x, u['g'].real.data, label=f"n={n}")
```
kburns, line 73: Might be good to do `(u['g']/np.max(u['g'])).real` to normalize the amplitude and phase of the eigenvectors. It doesn't make much difference here, but is good practice.

examples/evp_1d_waves_on_a_string/waves_on_a_string.py
```diff
@@ -67,3 +67,11 @@
...
+plt.xlim(0, 1)
+plt.legend(loc="lower right")
+plt.ylabel(r"mode structure")
+plt.xlabel(r"x")
```
kburns, line 77: Maybe add `plt.tight_layout()` and `plt.savefig("eigenvectors.pdf")` to save the plots.

kburns (changes requested):
Thanks! A few tweaks for it to run and save the plot when executed from the command line.

hdrake (commented):
Done, thanks!

PR #184, opened 2022-03-17
Enables merging of virtual files

### LINKED ISSUE #182: [d3] Merging virtual files into single files
Virtual files (and the partial files they read) are starting to cause me issues with my filesystem quotas on Pleiades. I've created a little script based on d2's merging logic that merges a virtual file into a single, merged dataset (attached here, in .txt format: [merge_virtual_files.py.txt](https://github.com/DedalusProject/dedalus/files/8227629/merge_virtual_files.py.txt)).

If there's interest, I can improve this and work towards a pull request to put this into d3's tools/post.py file.

evanhanders:
This PR fixes #182 by adding two functions to tools/post.py (merge_virtual_analysis(), merge_virtual_file()) and updating merge_setup() to allow virtual files to be merged into single files.

This file also brings merge_setup() from d2 -> d3, merging together the scales (which are now distributed and hashed, unlike in d2).

The workflow of the virtual file merge is:

1. Move virtual file from e.g., snapshots_s1.h5 -> tmp_snapshots_s1.h5
2. Create 'full' file from data in merged file called snapshots_s1.h5
3. Delete tmp_snapshots_s1.h5

The workflow of the full merge from partial files is identical to d2 (creates snapshots_s1.h5 from snapshots_s1/snapshots_s1_p*.h5).

@kburns and @jsoishi , it would be great to have both of you review this; I don't think I am allowed to assign reviewers.

### DIFF
--- a/dedalus/tests_parallel/test_output_parallel.py
+++ b/dedalus/tests_parallel/test_output_parallel.py
@@ -5,7 +5,7 @@ import pytest
 import numpy as np
 from dedalus.core import coords, distributor, basis, field, operators, problems, solvers, timesteppers, arithmetic
 from dedalus.tools.parallel import Sync
-from dedalus.tools.post import merge_virtual_analysis
+from dedalus.tools import post
 import shutil
 import h5py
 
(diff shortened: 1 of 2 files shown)


kburns:
Thanks!  Hmm it's a good question about how to name things.  Maybe what we should do is add something to the names of the virtual files when we initially create them, like `snapshots/snapshots_s1.vh5` or `snapshots/snapshots_s1v.h5` or `snapshots/snapshots_s1_virtual.h5`. And then the merged file could have the regular name `snapshots/snapshots_s1.h5`.  I like this because in a world where parallel hdf5 worked perfectly, we would just be directly writing to the `snapshots/snapshots_s1.h5` joint file without process files in the first place. Any thoughts @jsoishi?

jsoishi:
I agree that we should reserve snapshots/snapshots_s1.h5 for the truly
merged file because of the hope that someday we will have a functional
parallel hDF5 mechanism, and I think that changing the file extension would
be confusing, since .h5 is such an established standard.

What about snapshots/virt_snapshots_s1.h5? I'm afraid that snapshots_s1v.h5
might be too subtle...

kburns:
Maybe the merge script should just overwrite the virtual file, using the same name?  That way no plotting scripts, etc., downstream have to adjust to the change. I think this makes sense since the point of virtual files is to appear just like merged files, so in theory nothing in the workflow should care if the file is virtual or concrete. This would also avoid having both virtual and concrete merged files side-by-side, which could result in duplicated plots or similar if you're doing analysis like `python3 plot_file.py snapshots/*.h5`.

jsoishi:
Even better idea. +1

evanhanders:
This was actually the first implementation I tried (just replacing the virtual file with a merged file of the same name), but I was afraid this was a bit risky because it basically sets cleanup=True by default. Is that OK though?

I think there's two (almost identical) ways to make this happen. What's your preference?

1. move virtual file from snapshots_s1.h5 -> virt_snapshots_s1.h5. Merge it into snapshots_s1.h5. delete virt_snapshots_s1.h5 and snapshots_s1/.
2. merge snapshots_s1.h5 into merged_snapshots_s1.h5. delete snapshots_s1.h5 and snapshots_s1/. Move merged_snapshots_s1.h5 -> snapshots_s1.h5

kburns:
I think the first isn't too bad, and it doesn't have to (and probably shouldn't) delete the process files by default, just overwrite or move/delete the virtual merged file in place of the concrete merged file. We can also clean things up to make it easier to recreate the virtual merged file from the process files in post-processing, instead of just online.

lecoanet:
It would also be useful to have a merge that works without a virtual file.

kburns:
Marking as a draft since we're still WIP. I'm free the next few days/week if we want to push more on this.

evanhanders:
I'm happy to make a push ~tomorrow and early next week to get this together. Just to check, here's what we want to do, right?

1. Clean up the 'merged_' files that this does right now (as we discussed a few weeks ago
2. Get a merge that creates the virtual file working

Anything else?

evanhanders:
@kburns, @jsoishi, @lecoanet -- This PR has been updated with the requested changes:

1. Virtual files now merge 'in-place', so the end product of merging a virtual file called 'snapshots_s1.h5' is a merge file of the same name.
2. Full files can now be merged from partial files, and the distributed, hashed scales are properly merged.

kburns:
Thanks!

jsoishi:
BTW: some incoming changes to HDF5 files in d3 on my fork.

PR #185, opened 2022-03-23
S2 and shell latitudinal NCCs

kburns:
This PR includes work on S2 and shell latitudinal NCCs.


kburns:
Meridional NCCs and operators now seem to be working pretty well in the shell.  Problems will fail if there are both radial and meridional NCCs (currently caught by tests). This can be avoided now by promoting all NCCs to meridional bases, even if they are only radial. 

The PR also includes the sparse EVP solver, but this test is currently failing with some ARPACK convergence error -- haven't dug into this yet.

kburns:
@lecoanet anything else you think we should add for this PR?  Real variables would be nice, but I don't think should hold this up, since it's useful enough now for EVPs. Maybe we should fix it to allow radial NCCs and not require promotion to meridional for everything?

lecoanet:
I would like to add the rotating RBC problem to examples. Otherwise I think the PR is good. There are other things we could add, but nothing seems crucial.

lecoanet:
I added a rotating RBC example problem. I don't think a zcross operator is important. It is nice for S2 because it requires the skew operator which is confusing. I think it is more intuitive how we do this in the shell, so making a specific operator does not feel necessary to me. I would be fine merging in this PR at this stage.

kburns:
I'm just polishing the example script to match the syntax of the other examples -- but any reason not to use the first-order formulation here?  Looks like it gives very similar eigenvalues to the second-order.

lecoanet:
Go for the 1st order formulation -- I think I started with 1st order formulation, but switched to 2nd order when I was trying to find the discrepancy with the Dormy paper, and never switched back. I think it's a good idea to do 1st order for consistency with other example scripts.

PR #191, opened 2022-04-26
Truncation and conversion for scipy & FFTW DCTs

lecoanet:
When we take forward transforms with dealias scales > 1, the data gets truncated. Empirically, it seems to be advantageous to perform the truncation before applying any conversion operators (e.g., to the natural output basis of derivative operators). We have a DEALIAS_BEFORE_CONVERTING configuration option, but it was not implemented for scipy or FFTW DCTs. This pull request allows the user to apply the dealiasing truncation before applying the conversion matrix using this configuration option. We now pass all tests in the test_transforms.py script.

### DIFF
--- a/dedalus/core/transforms.py
+++ b/dedalus/core/transforms.py
@@ -816,9 +816,12 @@ class FastChebyshevTransform(JacobiTransform):
             self.resize_rescale_backward = self._resize_rescale_backward
         else:
             # Conversion matrices
-            self.forward_conversion = jacobi.conversion_matrix(self.N, a0, b0, a, b)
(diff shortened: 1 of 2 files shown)


kburns:
Looks good to me and passes all tests in the actions 👍.

kburns (approved):

bpbrown (approved):
Tested in a variety of 2-d Rayleigh-Benard simulations, and I'm seeing good and consistent results with this corrected dealiasing approach.  Also tested in much more complicated 2-d fully compressible equations, and saw good & consistent results there too.

Issue #190, opened 2022-04-25
HDF5 dims labels are incorrect for vector quantities


jsoishi:
Vector quantities output to HDF5 files are missing the dimension corresponding to vector/tensor component(s). Instead they currently skip that dimension, moving to the next one and leaving the trailing dimensions blank. Here's an example from a 2D velocity field. 

```python
u_dset = df['tasks/u']
for i in range(len(u_dset.shape)):
    print("dimension {} label: {}".format(i,u_dset.dims[i].label))
```

kburns:
With #196, the tensor axes are now skipped (no scales) so that the spatial scales are at least attached to the correct dimensions.

PR #192, opened 2022-04-26
Updates to HDF5 metadata

jsoishi:
- Fixes issue #190
- addes scale names to scale objects as well as tasks
- addes coordinate scales `coords` for tensor component dimensions populated from coordinate name strings from the CoordinateSystem object

### DIFF
--- a/dedalus/core/evaluator.py
+++ b/dedalus/core/evaluator.py
@@ -631,6 +631,29 @@ class FileHandler(Handler):
             scale.make_scale(sn)
             dset.dims[0].attach_scale(scale)
 
+        # Tensor components
+        skip = 1
+        for comp in op.tensorsig:
+            data = np.array(comp.names,dtype='S')
+            sn = 'coords'
+            if self.dist.comm_cart.rank == 0:
+                scale_hash = hashlib.sha1(data).hexdigest()
+            else:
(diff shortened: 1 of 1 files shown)


kburns:
Looks good. Only issue I can think of is that this could be misleading if we save vectors or tensors in coeff space, since there the coefficients may correspond to spin/regularity components, not the coordinate components like in grid space.

jsoishi:
That's a good point. We could do something like s_1/2, s_-1/2, r_0, r_1, etc in that case, basis dependent?

kburns:
Yeah or maybe just `{-1, 0, +1}` or even `{-, 0, +}`, and you have to know its spin or regularity based on the context?  Any thoughts @lecoanet?

lecoanet:
Our ordering is {-, +, 0}. Using {-, +, 0} for both spin and regularity bases would be fine. But {s-, s+, s0} and {r-, r+, r0} might be a bit more descriptive. Don't think we ever have spin 1/2.

jsoishi:
Too much QM teaching...

kburns:
@jsoishi want to push this to a separate branch on the main repo?

kburns:
Marking as a draft since we're still WIP.

Issue #189, opened 2022-04-24
User scripts can no longer control logger outputs to files


bpbrown:
In the past, it has been very handy to have user scripts specify logger output to files.  

With a code block like this:
```python
from dedalus.tools.config import config
config['logging']['filename'] = os.path.join(data_dir,'logs/dedalus_log')
config['logging']['file_level'] = 'DEBUG'

import dedalus.public as de
```
a user script will output debug level loggers to a a set of files in `data_dir/logs/dedalus_log_p#.log`, with `data_dir` a potentially script-specific location (e.g., a directory named with the parameters of a specific experiment).  Crucially, in this code block, `config` is imported from `dedalus.tools.config`, and relevant settings are made, before `dedalus.public` is imported.  `dedalus.public` in the past imported `from .tools import logging as _logging_setup`.

With commit 905fcff9a5c72fb18d1e28a367af8ec96150c7c5, this has changed, and the earlier implementation specifying logger output to files within user scripts now no longer works.

The key change appears to be the removal of 
```python
from .tools import logging as _logging_setup
```
from `public.py` and the inclusion of 
```python
# Import custom logging to setup rootlogger
from .tools import logging as _logging_setup
```
in `__init__.py`.

I have confirmed that user script logging control works in immediately preceding commit 030fa13cf2dddd2eb1c3c71ff7ce3b85125d3531, and I've tested the current master 76955c8df062e15164e73b38cf6d3a45a28a46e8, plus I have reverted these changes on the current master commit to confirm they are the source of the changed behaviour.

It's not clear to me how to solve this problem; adding the import from `.tools` block back into `public.py` still fails, and the only approach that has worked in my testing is to revert the changes in `__init.py__` (remove the import from `.tools`) and `public.py` (add the import from `.tools`) simultaneously.  But `__init__.py` needs this import to have a correctly performing logging for the threading warning.

bpbrown:
MWE:

```python
import os

import logging
logger = logging.getLogger(__name__)

data_dir = 'test_case'

logger.info("this is a test script, writing to: {:}".format(data_dir))

from dedalus.tools.config import config
config['logging']['filename'] = os.path.join(data_dir,'logs/dedalus_log')
config['logging']['file_level'] = 'DEBUG'

if not os.path.exists('{:s}/'.format(data_dir)):
    os.mkdir('{:s}/'.format(data_dir))
logdir = os.path.join(data_dir,'logs')
if not os.path.exists(logdir):
[...]

PR #195, opened 2022-04-28
Update logging setup for easier script-based manipulation

kburns:
This addresses #189 by adding a helper function for creating a logging file after dedalus is instantiated/imported. This is needed because we now log from the top-level __init__.py to check threading environment variables.


bpbrown (approved):
Tested and this is a vast improvement.  Works solidly, removes 10+ lines of work-around code in individual user scripts.

PR #196, opened 2022-05-04
Restructure hdf5 output handlers

kburns:
This PR restructures the FileHandler class into several classes with different approaches to handing hdf5 outputs in parallel, including:
1) A gather-based approach where just the root node writes data
2) Parallel HDF5 writes
3) Individual process files and a virtual merged dataset, where the process files have been stripped down (scales removed, etc.).

The motivation is to eliminate the need for merge scripts in most cases, and to simplify the code in the FileHandler class.

### DIFF
--- a/dedalus/core/evaluator.py
+++ b/dedalus/core/evaluator.py
@@ -1,7 +1,4 @@
-"""
-Class for centralized evaluation of expression trees.
-
-"""
+"""Class for centralized evaluation of expression trees and handling the results."""
 
 import os
 import re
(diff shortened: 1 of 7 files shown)


kburns:
@jsoishi did my best to make an mpio output class, but it (mostly) doesn't work so for now I just did my best to categorize and document the various errors I'm seeing. That just leaves redoing the virtual output class.

kburns:
I think everything (gather, mpio, and virtual file handlers) should be working now. Would be great to stress test these a bit!

PR #204, opened 2022-06-09
Modified Product to work for NCC's with no bases.

lecoanet:
I use the _last_axis_field_ncc_matrix method from the Basis base class if the NCC and operand have no basis.


lecoanet:
Added test problem to solve algebraic equation: v*u = F, where v=-1 and F=-3. This test passes after commit a1900a0.

kburns:
Tests passing

PR #207, opened 2022-06-22
Renamed subdirectory in the example directory

carbonatezero:
Rename examples/**lbvp**_2d_poisson/poisson.py to examples/**ibvp**_2d_poisson/poisson.py


carbonatezero:
Sorry. The original one is correct. I should cancel this PR.

PR #214, opened 2022-07-11
Change format when referring to coordinates

iuryt:


kburns:
Thanks for finding the typo! I think it's better to leave the coordinates in plain text rather than latex here, though, because the compound names like `kx` and `nz` don't render well in latex.

PR #215, opened 2022-07-11
Remove empty cell and fix a typo

iuryt:


kburns:
Thanks for finding the empty cell!  The text was correct before, however.  This is pointing out that the vector/tensor components are different in coefficient space versus grid space, depending on whether you're in cartesian or curvilinear coordinates.

Issue #218, opened 2022-07-21
Support subproblem selection in LBVP


kburns:
In some cases, we might want to solve LBVPs only over certain subproblems, as we do for EVPs. It would be nice to be able to to do this and add lazy construction of the LBVP matrices, as for EVPs.

kburns:
Closed with #222.

PR #222, opened 2022-08-17
Solver optimizations and cleanup

kburns:
This PR includes a few solver optimizations, including:
* Combining matvecs for the subsystems in the same subproblem
* Combining linear solvers for the subsystems in the same subproblem

and a lot of solver class cleanup and rounding out a few features, including:
* Adding v2-style left eigenvector support the sparse EVP
* Adding `rebuild_matrices` options to the LBVP/EVP to support on-the-fly field NCC modifications
* Allowing you to specify specific subsystems for the LBVP like for the EVP (addressing #218).

It would be nice to add some left eigenvector tests / examples.

### DIFF
--- a/dedalus/core/problems.py
+++ b/dedalus/core/problems.py
@@ -1,27 +1,15 @@
-"""
-Classes for representing systems of equations.
+"""Classes for representing systems of equations."""
 
-"""
-
-from collections import OrderedDict
 import numpy as np
-from mpi4py import MPI
-from functools import reduce
 from collections import ChainMap
 
-from . import field
 from .field import Operand, Field
-from . import future
 from . import arithmetic
 from . import operators
 from . import solvers
 from ..tools import parsing
-from ..tools.cache import CachedAttribute
 from ..tools.general import unify_attributes
-from ..tools.exceptions import SymbolicParsingError
 from ..tools.exceptions import UnsupportedEquationError
-
-from ..tools.config import config
 
 import logging
 logger = logging.getLogger(__name__.split('.')[-1])
@@ -45,104 +33,8 @@ parseables.update({name: getattr(arithmetic, name) for name in arithmetic.__all_
 parseables.update(arithmetic.aliases)
 
 
-class Namespace(OrderedDict):
-    """Class ensuring a conflict-free namespace for parsing."""
-
-    __slots__ = 'allow_overwrites'
-
-    def __init__(self):
-        super().__init__()
-        self.allow_overwrites = True
-
-    def __setitem__(self, key, value):
-        if key in self:
-            if not self.allow_overwrites:
-                raise SymbolicParsingError("Name '{}' is used multiple times.".format(key))
-        else:
-            if not key.isidentifier():
-                raise SymbolicParsingError("Name '{}' is not a valid identifier.".format(key))
-        super().__setitem__(key, value)
-
-    def copy(self):
-        """Copy entire namespace."""
-        copy = Namespace()
-        copy.update(self)
-        copy.allow_overwrites = self.allow_overwrites
-        return copy
-
-    def add_substitutions(self, substitutions):
-        """Parse substitutions in current namespace before adding to self."""
-        for call, result in substitutions.items():
-            # Convert function calls to lambda expressions
-            head, func = parsing.lambdify_functions(call, result)
-            # Evaluate in current namespace
-            self[head] = eval(func, self)
-
-
-# class Equation:
-
-#     def __init__(self, eq_string, condition):
-#         self.eq_string = eq_string
-#         self.condition = condition
-
-#         self.LHS_string
-#         self.RHS_string
-#         self.LHS_object
-#         self.RHS_object
-
-
-    # @property
-    # def bases(self):
-    #     return self.LHS.bases
-
-    # @property
-    # def subdomain(self):
-    #     return self.LHS.subdomain
-
-    # @property
-    # def separability(self):
-    #     return self.LHS.separability
-
-    # def check_condition(self, group_dict):
-    #     pass
-
-
-
 class ProblemBase:
-    """
-    Base class for problems consisting of a system of PDEs, constraints, and
-    boundary conditions.
-
-    Parameters
-    ----------
-    domain : domain object
-        Problem domain
-    variables : list of str
-        List of variable names, e.g. ['u', 'v', 'w']
-
-    Attributes
-    ----------
-    parameters : OrderedDict
-        External parameters used in the equations, and held constant during integration.
-    substitutions : OrderedDict
-        String-substitutions to be used in parsing.
-
-    Notes
-    -----
-    Equations are entered as strings of the form "LHS = RHS", where the
-    left-hand side contains terms that are linear in the dependent variables
-    (and will be parsed into a sparse matrix system), and the right-hand side
-    contains terms that are non-linear (and will be parsed into operator trees).
-
-    The specified axes (from domain), variables, parameters, and substitutions
-    are recognized by the parser, along with the built-in operators, which
-    include spatial derivatives (of the form "dx()" for an axis named "x") and
-    basic mathematical operators (trigonometric and logarithmic).
-
-    The LHS terms must be linear in the specified variables and first-order in
-    coupled derivatives.
-
-    """
+    """Base class for all problem types."""
 
     def __init__(self, variables, namespace=None):
         self.variables = variables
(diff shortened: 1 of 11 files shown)


kburns:
Added big clean up of the problem classes, including adding back in all equation condition checking and sensible error messages when problems are detected.

dedalus/core/problems.py
```diff
@@ -45,281 +33,134 @@
...
 
     def __init__(self, variables, namespace=None):
         self.variables = variables
         self.LHS_variables = variables
         self.dist = unify_attributes(variables, 'dist')
         self.equations = self.eqs = []
         # Build namespace via chainmap to react to upstream changes
-        # Priority: local dict, specified namespace, parseables
+        # Priority: local namespace, external namespace, parseables
+        self.local_namespace = {}
```
lecoanet, line 46: Why do we make a local_namespace? Is this so someone could modify this after the problem instantiation? I don't see this used anywhere right now.

lecoanet:
Should we rewrite the RBC EVP example to use the rebuild_matrices? It is a bit complicated because kx isn't a parameter, it is part of the field. Maybe we could make it a 1D problem and define a custom grad operator? Sounds unnecessarily complicated....

dedalus/core/problems.py
kburns: This is so that you can directly add things to the namespace after instantiating the problem without polluting locals, i.e. `problem.namespace['a'] = A` will make `a` available for equation parsing, but won't create/change `a` in the script namespace.

kburns:
Yeah it would be great to have a better way to handle that eventually.  Maybe once we have domain remapping / metric tensors implemented, you could just modify the horizontal stretching parametrically and rebuild matrices.  But for now this is probably the cleanest thing besides writing the equations scalar/component style like in v2.

lecoanet:
I tested performance on an Nr=256, Lmax=127 problem on Pleiades with 1024 cores (mesh=[32, 32]). There is some run-to-run variability which I think is due to differences in parallel transpose times. The new PR does not affect performance within the run-to-run variability. However, I looked at how much time is actually spent on the matrix solve and the csr_matvec/csr_matvecs. I measured these relative to the amount of time spent on attempt_tasks, which I think should be fairly consistent across simulations because it does not depend on the transpose timings.

The new matrix solves are very slightly faster, by about 2%. That's to say that if before the solves were 3% of the simulation time, now they are 2.94% of the simulation time. However, the new csr_matvecs are more robustly slower, by about 30%. They've gone from ~1.5% of the simulation time to ~1.95% of the simulation time. Overall, the simulations are slower with the changes made in this PR.

dedalus/core/solvers.py
```diff
@@ -513,7 +516,7 @@ def __init__(self, problem, timestepper, enforce_real_cadence=100, warmup_iterat
         # Initialize timestepper
         self.timestepper = timestepper(self)
         # Attributes
-        self.sim_time = self.initial_sim_time = 0  # TODO: allow picking up from current problem sim time?
+        self.sim_time = self.initial_sim_time = problem.time['g'].ravel()[0]
```
lecoanet, line 519: This doesn't work in parallel.

lecoanet:
More confusion on performance... I tried testing the internally heated convection problem on 4 cores on my laptop. There it seems like the new solves are faster by ~15% and csr_matvecs are faster by a factor of 2. However, running the problem on the PR is consistently slower (by 0.5%) than on master. Digging down, I noticed a big difference in the amount of time spent in "concatenate" call on increment_group and decrement_group. On master, this takes 4.79% of the simulation time, whereas on the PR it takes 12.79% of the simulation time. The weird thing is there doesn't seem to be any difference in concatenate timing on Pleiades. Could there be something about data allocation/ordering in the new gather/scatter that makes the concatenate call on increment_group and decrement_group slower?

dedalus/core/solvers.py
kburns: Fixed

kburns:
I'm not sure, maybe it's interacting with load balancing in some strange way? Those methods are directly concatenating all local data, not going through the subsystem/subproblem gather or scatter interfaces.

lecoanet:
I've now also tested performance on my desktop (mac studio with new mac chip). On the internally heated convection problem (on 4 cores), I find the new csr_matvecs are _faster_ than the implementation on master by ~x2, similar to my laptop. The solves are about the same speed. Overall the simulation seems to be ~3% faster on the PR, although there is some run-to-run variability. I also tested the 2D rayleigh benard problem (on 4 cores). There I found the csr_matvecs are _slower_ than the implementation on master by ~x2; the solves are similar. Mostly due to the slow matvecs's, the 2D RBC on the PR were 15% slower than on master.

To summarize the results of these various tests: It seems like the new matrix solves are similar in performance to the old matrix solves. The new matvecs's are sometimes much faster, sometimes much slower than the old ones. If we want to improve performance, I would suggest that we make a configuration option where the user can choose whether to use individual matvec's for each subsystem, or do one big matvecs for the whole subproblem.

kburns:
Seems very non-reproducible. On my mac laptop, the 2D RBC example is 15% faster on the PR than on master, entirely because csr_matvecs is 3x faster than csr_matvec.

dedalus/core/problems.py
```diff
@@ -329,201 +170,165 @@ def _build_matrix_expressions(self, eqn):
...
 
-    @CachedAttribute
-    def namespace_additions(self):
-        """Build namespace for problem parsing."""
-        # Variable perturbations
-        self.perturbations = [Field(bases=var.bases, name='δ'+var.name, dtype=var.dtype) for var in self.variables]
-        return {pert.name: pert for pert in self.perturbations}
+    def _check_equation_conditions(self, eqn):
+        """Check equation conditions."""
+        # No conditions
```
lecoanet, line 231: Do we need to check linearity of LHS?

lecoanet: Maybe also domain containment?

dedalus/core/solvers.py
```diff
@@ -144,99 +138,112 @@ class EigenvalueSolver(SolverBase):
...
-
+        subproblem : Subproblem object
+            Subproblem for which to solve the EVP.
+        rebuild_matrices : bool, optional
+            Rebuild LHS matrices if coefficients have changed (default: False).
+        left : bool, optional
+            Solve for the left eigenvectors of the system in addition to the
+            right eigenvectors. The left eigenvectors are the right eigenvectors
+            of the conjugate-transposed problem. Follows same definition described
+            in scipy.linalg.eig documentation (default: False).
```
lecoanet, line 180: Should we mention we're multiplying by -M? I don't think that is part of the scipy.linalg.eig documentation.

dedalus/core/problems.py
kburns: We don't need them since the current NLBVP setup takes `LHS = RHS` and converts it to a straight nonlinear root-finding using both sides: `H(X) = LHS - RHS = 0`, so there's no restrictions on LHS vs RHS.

lecoanet:
On Pleiades, the 2D RBC problem on 4 cores is ~5% faster on master than on the PR. I think most of the difference is csr_matvecs takes 50% more time on the PR. I don't have any evidence that the performance is non-reproducible, but it does seem to be different on different problems and different machines. For many of the cases I've tested, it seems to make things slower, so I don't think we should remove the old implementation from the code.

kburns:
Can you tell from your examples if the speed differences are in the `_sparsetools.csr_matvec(s)` or just `array.csr_matvec(s)`?  If the former, then I think we can just put the fallback option inside of `array.csr_matvecs` so we don't have to pollute all the solver/timestepper code with extra subsystem loops.

lecoanet (commented):
I've gone through the PR and it looks good. The error messages around equation entry are now much more expressive!! There is still an opaque error message that can come up: when you accidentally add a scalar to a vector. You get an error saying "ValueError: Objects are not all equal." which may be hard to interpret. I guess it is not specifically an issue with equation entry, as it can happen when you form any expression. But I think it would be useful to make this clearer. Otherwise I think it is good to merge after we resolve the performance issues with the matvec's.

lecoanet:
The speed differences appear to be in _sparsetools.csr_matvec(s). I think you're right that putting the options in array.csr_matvecs might be a good solution.

kburns:
@lecoanet does it look like this config option enough to get the pre-PR performance on your machines?

lecoanet:
I tested the 2D RBC problem on Pleiades with 4 cores. Before the _sparsetools.csr_matvecs was spending ~50% more time than the _sparsetools.csr_matvec calls on the master branch, but the time spent in array's csr_matvec(s) calls were similar. With SPLIT_CSR_MATVECS = True, the same amount of time is spent in _sparetools.csr_matvec as on master, but now significantly more time in spent in array's csr_matvec. Overall, the extra time spent in array means that slightly more time is spent on the matvec's with SPLIT_CSR_MATVECS = True. That means the matvec's are still ~50% slower on the PR than on master, even when using the configuration option. I can check to see if it helps things at all on my desktop when I get back to Chicago in a couple of weeks.

kburns:
Do you mean array.csr_matvecs? With SPLIT_CSR_MATVECS on the PR, array.csr_matvec shouldn't be called at all, it should directly call _sparsetools.csr_matvec from array.csr_matvecs.

kburns:
Also what does the difference in overall runtimes (without profiling) look like now?  The differences between master and with/without splitting on the branch are now lower than the run-to-run variations on both my linux desktop and the flatiron clusters.

kburns:
I think we should just merge this in for now since we have the config option, and there's lots of other nice API things we'd like to continue building on, but happy to work on more performance analysis and any necessary fixes in another PR.

jsoishi:
+1

PR #224, opened 2022-09-01
better error message for adding sphere basis with different raddi

### LINKED ISSUE #216: [d3] Error message when adding/multiplying S2 bases with different radii is not helpful
If you accidentally try to add two S2 bases with different radii, the error message is 

`TypeError: unsupported operand type(s) for +: 'SphereBasis' and 'SphereBasis'`

This does not explain the nature of the error. Given this is a very easy error to make, particularly when setting boundary conditions in spherical shells, it would be good to at least make a suggestion of checking radii.

jayceslesar:
closes #216


jayceslesar:
Probably other places we could add these so let me know

kburns:
Thanks!

Issue #203, opened 2022-06-08
Bug in basis caching based on multidimensional vs scalar dealias keyword


kburns:
Looks like passing a scalar for `dealias` for a multidimensional basis will be correctly expanding, but only after caching, causing actually identical bases to be instantiated separately.

Ref: https://groups.google.com/g/dedalus-users/c/X0qr_S4sxzk/m/QzIsWtcMCAAJ

kburns:
Fixed in 65ab92d8d63a2b4f10e4f7b4aec8c340158807cc.

PR #225, opened 2022-10-05
Annulus improvements

kburns:
Adds some fixes to get the annulus working. Includes centrifugal convection example.


kburns:
Ok to get the annulus working, we added a cache preprocessing step to resolve defaults and normalize the types and lengths of all arguments to the user-facing basis classes.  This should also fix #203, but I need to double-check that.  

The standard test suite is passing, but this doesnt have many NCC tests I think, so it would be good to also test a few other scripts.

PR #227, opened 2022-10-08
GeneralFunction capabilities for d3

lecoanet:
This is a first attempt to get GeneralFunction working in d3. I have used it to add stochastic forcing to an equation. Right now the user needs to pass in a distributor, domain, tensorsig and dtype -- normally these would come from the operand, but the GeneralFunction does not necessarily have any operand! I'm not sure if there is any better way to get these. I also need to make unit tests (and make sure it works for more sophisticated functions!).


kburns:
Sorry for the delay. Is this good to go as far as we know?

kburns (approved):

PR #238, opened 2023-01-30
Output support for BVPs

kburns:
This PR will have some cleanups for evaluators & outputs, particularly supporting outputting from BVPs to HDF5.

### DIFF
--- a/dedalus/core/evaluator.py
+++ b/dedalus/core/evaluator.py
@@ -247,10 +247,16 @@ class Handler:
                 self.last_wall_div = wall_div
         # Sim time
         if self.sim_dt:
-            sim_div = kw['sim_time'] // self.sim_dt
-            if sim_div > self.last_sim_div:
-                scheduled = True
-                self.last_sim_div = sim_div
+            # Output if the output target closest to the current time hasn't triggered
+            # an output yet, and the next timestep will not bring you closer.
+            t = kw['sim_time']
+            dt = kw['timestep']
+            closest_sim_div = int(np.round(t / self.sim_dt))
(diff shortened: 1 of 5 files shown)


kyle-augustson:
Hey all,

I have been instantiating an evaluator class for the eigenvalue solver class outside of the EigenvalueSolver class in core/solvers.py since it is not yet implemented there.  In my scripts, rather, I have been doing something that seems to be done for all the solver classes except EVPs:

```python
namespace = {}
solver.evaluator = evaluator.Evaluator(solver.dist, namespace)
```

For the other solver classes, this evaluator is then often to evaluate the right-hand-side terms, which isn't necessary for the EVPs and perhaps why it was omitted in the EVPs. 

 In any case, since all the classes seem to use this in their initialization, would it be possible to move this to the `SolverBase` super class?  Or at least add

```python
namespace = {}
self.evaluator = Evaluator(self.dist, namespace)
```

to the `EigenvalueSolver` class initialization?

Cheers,
Kyle

kburns:
Ok added the evaluator to the base class, and made some other simple cleanups. I think outputs should now work with all the solver classes.

kburns:
The NLBVP, LBVP, and IVP now all have `solver.evaluate_handlers` methods to directly trigger rights. Maybe we can add something for the EVP too, but I guess it's less clear what sequential writes mean there -- there's not really "iterations" necessarily.  You could use it manually to save a sequence of eigenmodes, I guess. But we'd probably want some additional interface to save the eigenvalues too, or something.  Maybe that's beyond scope for this PR.

Issue #198, opened 2022-05-16
2.5D cartesian problems not working


kburns:
Ref: https://groups.google.com/g/dedalus-users/c/YKAuPqh5Q0M/m/7u7rdSTQAwAJ?utm_medium=email&utm_source=footer

bpbrown:
This is a general and larger problem for "half dimensional" problems in cartesian coordinates.  

Half dimensions are coordinate dimensions which do not have bases associated with them, e.g. the following 1.5D system:
```python
coords = de.CartesianCoordinates('x', 'y', 'z')
dist = de.Distributor(coords, dtype=dtype)
zb = de.ChebyshevT(coords.coords[2], size=nz, bounds=(0, Lz), dealias=dealias)

A = dist.VectorField(coords, name='A', bases=zb)
```
where `x` and `y` are half dimensions, or 2.5D systems like this:
```python
coords = de.CartesianCoordinates('x', 'y', 'z')
dist = de.Distributor(coords, dtype=dtype)
ybasis = de.ComplexFourier(coords['y'], size=Ny, bounds=(0, Ly), dealias=dealias)
zbasis = de.ComplexFourier(coords['z'], size=Nz, bounds=(0, Lz), dealias=dealias)

bases = (ybasis, zbasis)

A = dist.VectorField(coords, name='A', bases=bases)
```
where `x` is a half dimension.

Specific problems that we know of:

- [ ]  when attempting to use `de.Gradient()` or similar operators in half dimensions, an `AttributeError: 'int' object has no attribute 'prep_nccs'` is raised.
- [ ]  when attempting to use `dedalus.extras.flow_tools.CFL()`, an `AttributeError: 'NoneType' object has no attribute 'dealias'`


We are working on fixes to this in the linked `d3-half-dimensions` branch.

bpbrown:
The first problem:

- when attempting to use de.Gradient() or similar operators in half dimensions, an AttributeError: 'int' object has no attribute 'prep_nccs' is raised.

comes about because in the half-dimension direction, the derivative is zero.  Normally, we want to drop zeros from the system, but in this case we want to retain that zero as a field.  The current solution is to check for zeros, by hand create a multiplication operation, and then change the multiplication value to zero manually:
```python
if args[i] == 0:
     args[i] = 2*operand
     args[i].args[0] = 0
```
Note that approaches like:
```python
from arithmetic import Multiply
if args[i] == 0:
     args[i] = Multiply(0, operand)
```
[...]

PR #239, opened 2023-02-25
D3 half dimensions

bpbrown:
This PR adds half-dimensions (dimensions with a coordinate direction, but no associated basis) and addresses functionality requested in #198.

### DIFF
--- a/dedalus/core/basis.py
+++ b/dedalus/core/basis.py
@@ -6073,9 +6073,10 @@ class CartesianAdvectiveCFL(operators.AdvectiveCFL):
         spacing = []
         for i, c in enumerate(coordsys.coords):
             basis = velocity.domain.get_basis(c)
-            dealias = basis.dealias[0]
-            axis_spacing = basis.local_grid_spacing(i, dealias) * dealias
-            N = basis.grid_shape((dealias,))[0]
+            if basis:
+                dealias = basis.dealias[0]
+                axis_spacing = basis.local_grid_spacing(i, dealias) * dealias
+                N = basis.grid_shape((dealias,))[0]
             if isinstance(basis, Jacobi) and basis.a == -1/2 and basis.b == -1/2:
                 #Special case for ChebyshevT (a=b=-1/2)
                 local_elements = basis.dist.grid_layout.local_elements(basis.domain, scales=dealias)[i]
@@ -6086,6 +6087,8 @@ class CartesianAdvectiveCFL(operators.AdvectiveCFL):
                 #Special case for Fourier
                 native_spacing = 2 * np.pi / N
                 axis_spacing[:] = dealias * native_spacing * basis.COV.stretch
+            elif basis is None:
+                axis_spacing = np.inf
             spacing.append(axis_spacing)
         return spacing
 
--- a/dedalus/core/operators.py
+++ b/dedalus/core/operators.py
@@ -2307,7 +2307,6 @@ class CartesianGradient(Gradient):
             if args[i] == 0:
                 args[i] = 2*operand
                 args[i].args[0] = 0
-        print(args)
         bases = self._build_bases(*args)
         args = [convert(arg, bases) for arg in args]
         LinearOperator.__init__(self, *args, out=out)
--- a/dedalus/core/operators.py
+++ b/dedalus/core/operators.py
@@ -2303,6 +2303,11 @@ class CartesianGradient(Gradient):
 
     def __init__(self, operand, coordsys, out=None):
         args = [Differentiate(operand, coord) for coord in coordsys.coords]
+        for i in range(len(args)):
+            if args[i] == 0:
+                args[i] = 2*operand
+                args[i].args[0] = 0
+        print(args)
         bases = self._build_bases(*args)
         args = [convert(arg, bases) for arg in args]
         LinearOperator.__init__(self, *args, out=out)
--- a/examples/evp_1d_rayleigh_benard/rayleigh_benard_evp.py
+++ b/examples/evp_1d_rayleigh_benard/rayleigh_benard_evp.py
@@ -28,58 +28,51 @@ import dedalus.public as d3
 import logging
 logger = logging.getLogger(__name__)
 
-first_order_form = False
 
 def max_growth_rate(Rayleigh, Prandtl, kx, Nz, NEV=10, target=0):
     """Compute maximum linear growth rate."""
 
     # Parameters
     Lz = 1
+    # Build Fourier basis for x with prescribed kx as the fundamental mode
+    Nx = 2
+    Lx = 2 * np.pi / kx
 
     # Bases
     coords = d3.CartesianCoordinates('x', 'z')
     dist = d3.Distributor(coords, dtype=np.complex128, comm=MPI.COMM_SELF)
+    xbasis = d3.ComplexFourier(coords['x'], size=Nx, bounds=(0, Lx))
     zbasis = d3.ChebyshevT(coords['z'], size=Nz, bounds=(0, Lz))
 
     # Fields
     omega = dist.Field(name='omega')
-    p = dist.Field(name='p', bases=zbasis)
-    b = dist.Field(name='b', bases=zbasis)
-    u = dist.VectorField(coords, name='u', bases=zbasis)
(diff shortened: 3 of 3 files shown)


bpbrown:
The issue raised in #198 is a general and larger problem for "half dimensional" problems in cartesian coordinates.  

Half dimensions are coordinate dimensions which do not have bases associated with them, e.g. the following 1.5D system:
```python
coords = de.CartesianCoordinates('x', 'y', 'z')
dist = de.Distributor(coords, dtype=dtype)
zb = de.ChebyshevT(coords.coords[2], size=nz, bounds=(0, Lz), dealias=dealias)

A = dist.VectorField(coords, name='A', bases=zb)
```
where `x` and `y` are half dimensions, or 2.5D systems like this:
```python
coords = de.CartesianCoordinates('x', 'y', 'z')
dist = de.Distributor(coords, dtype=dtype)
ybasis = de.ComplexFourier(coords['y'], size=Ny, bounds=(0, Ly), dealias=dealias)
zbasis = de.ComplexFourier(coords['z'], size=Nz, bounds=(0, Lz), dealias=dealias)

bases = (ybasis, zbasis)

A = dist.VectorField(coords, name='A', bases=bases)
```
where `x` is a half dimension.

Specific problems that we know of:

- [x]  when attempting to use `de.Gradient()` or similar operators in half dimensions, an `AttributeError: 'int' object has no attribute 'prep_nccs'` is raised.
- [x]  when attempting to use `dedalus.extras.flow_tools.CFL()`, an `AttributeError: 'NoneType' object has no attribute 'dealias'`

bpbrown:
The first problem:

- when attempting to use de.Gradient() or similar operators in half dimensions, an AttributeError: 'int' object has no attribute 'prep_nccs' is raised.

comes about because in the half-dimension direction, the derivative is zero.  Normally, we want to drop zeros from the system, but in this case we want to retain that zero as a field.  The current solution is to check for zeros, by hand create a multiplication operation, and then change the multiplication value to zero manually:
```python
if args[i] == 0:
     args[i] = 2*operand
     args[i].args[0] = 0
```
Note that approaches like:
```python
from arithmetic import Multiply
if args[i] == 0:
     args[i] = Multiply(0, operand)
```
fail because `Multiply()` itself triggers a `SkipDispatchException(output=0)` on zero values, and that is the generally desired behaviour.

bpbrown:
The second problem:

- when attempting to use `dedalus.extras.flow_tools.CFL()`, an `AttributeError: 'NoneType' object has no attribute 'dealias'`

comes about because
```python
basis = velocity.domain.get_basis(c)
```
returns a `NoneType` for basis for half-dimension coordinate `c`.

The current solution is to test on basis before trying to access `basis.dealias` and related features.  If `basis is None`, then a spacing of `np.inf` is returned.

bpbrown:
### Testing

We need a good test for this.  

Right now, as a starting point, I'm testing against a known 2.5D eigenvalue problem for kinematic dynamos with "Roberts flows of type 1".  Scripts for this problem are available in [bpbrown/kinematic_dynamos](https://github.com/bpbrown/kinematic_dynamos).  The relevant original literature is:

> Roberts, G.O., 1972,
> ``Dynamo action of fluid motions with two-dimensional periodicity'',
> Philosophical Transactions of the Royal Society of London

This problem has 2-D nccs in `y` and `z`, and has a half dimension in `x`.  
```python
coords = de.CartesianCoordinates('x', 'y', 'z')
dist = de.Distributor(coords, dtype=dtype)
ybasis = de.ComplexFourier(coords['y'], size=Ny, bounds=(0, Ly), dealias=dealias)
zbasis = de.ComplexFourier(coords['z'], size=Nz, bounds=(0, Lz), dealias=dealias)
```
Interestingly, in this problem, the half-dimension in `x` has modal structure with eigenfunctions `np.exp(1j*kx*x)`.  This means means the `dx()` derivatives are `1j*kx` rather than zero.

The scripts in this repo function on `master` if we define all gradient operators by hand:
```python
dx = lambda A: 1j*kx*A
dy = lambda A: de.Differentiate(A, coords['y'])
dz = lambda A: de.Differentiate(A, coords['z'])

div = lambda A:  dx(A@ex) + dy(A@ey) + dz(A@ez)
grad = lambda A: dx(A)*ex + dy(A)*ey + dz(A)*ez
lap = lambda A: dx(dx(A)) + dy(dy(A)) + dz(dz(A))
curl = lambda A: (dy(A@ez)-dz(A@ey))*ex + (dz(A@ex)-dx(A@ez))*ey + (dx(A@ey)-dy(A@ex))*ez
```

With `d3-half-dimensions`, we can use the built in operators, adding components for the half-dimension `x`, which handles the modal `x` structure:
```python
dx = lambda A: 1j*kx*A
div = lambda A:  de.div(A) + dx(A@ex)
grad = lambda A: de.Gradient(A, coords) + dx(A)*ex
lap = lambda A: de.lap(A) + dx(dx(A))
curl = lambda A: de.Curl(A) - dx(A@ez)*ey + dx(A@ey)*ez
```

We can test this system by looking at the peak growth rate for particular parameter values, by running:
```bash
python roberts_flow_peak.py --N=64
```

On `master` (bpbrown/kinematic_dynamos@e599450e6d4c377b40dcd5a78cc9bf26bb70ffcf), the critical value is:
```
ω = 0.17153311268155436 at kx = 0.5384618769875998
```
while on `d3-half-dimensions` (bpbrown/kinematic_dynamos@2d7db66fca75d48e7ebbfb8abf9977ea9a978909) the critical value is:
```
ω = 0.17153311268155444 at kx = 0.5384618717553397
```
which gives 15 digits of agreement for `ω` and 8 digits of agreement for `kx` between the two approaches; these are both almost certainly set by the tolerances in our `kx` search using `scipy.optimize.minimize()`.   These solve for the system with `flow=1, lambda=1/8` .

Other tests also show good agreement, and the resulting growth curves, computed with
```bash
python roberts_flow_2.5D_scan_lambda.py
```
agree in detail (by eye) with Robers 1972, figure 7.

bpbrown:
Testing with the 1-d Rayleigh-Benard onset EVP in `evp_1d_rayleigh_benard/rayleigh_benard_evp.py`, converted to use half-dimensions, reveals an interesting problem:
- in second order form, the computed growth curve matches expectations; at Rayleigh > critical, some modes are growing (positive growth rate) and some are decaying (negative growth rate)
- in first order form, the computer growth curve does not match expectations, though initially it looks fine (is not clearly spurious without a comparison curve).  But it's clearly wrong, as at Rayleigh > critical all modes are decaying.

To toggle between first and second order form, use the boolean:
```python
first_order_form = False
```
We need to understand this issue before proceeding with this PR, as choice of problem formulation should not lead to a change in basic answer, like this does.

bpbrown:
> Testing with the 1-d Rayleigh-Benard onset EVP in `evp_1d_rayleigh_benard/rayleigh_benard_evp.py`, converted to use half-dimensions, reveals an interesting problem:
> 
> * in second order form, the computed growth curve matches expectations; at Rayleigh > critical, some modes are growing (positive growth rate) and some are decaying (negative growth rate)
> * in first order form, the computer growth curve does not match expectations, though initially it looks fine (is not clearly spurious without a comparison curve).  But it's clearly wrong, as at Rayleigh > critical all modes are decaying.
>
> ...
> 
> We need to understand this issue before proceeding with this PR, as choice of problem formulation should not lead to a change in basic answer, like this does.

This issue originated in the local definitions of `grad` and `div`.

The original, incorrect form used:
```python
    grad = lambda A: d3.grad(A) + dx(A)*ex
    div = lambda A: d3.div(A) + dx(A@ex)
```
the correct form uses:
```python
    grad = lambda A: d3.grad(A) + ex*dx(A)
    div = lambda A: d3.div(A) + dx(ex@A)
```
the difference is between right (incorrect) and left (correct) multiplication and dot products between `A` and `ex`.

kburns:
I think we want the changes that allow solving IVPs and BVPs when the solution is truly lower-dimensional than the domain, but I don't think we necessarily want to promote this approach for EVPs, including the RBC example. My thinking is that for the EVP, you need additional boilerplate (redefining the multidimensional calculus operators) to implicitly implement a horizontal Fourier discretization. So instead, it's probably better to just use the Fourier basis and the built-in multidimensional operators.

PR #241, opened 2023-03-16
Update dedalus_tutorial_fields_operators.ipynb

carbonatezero:
Fixed an error in the notebook:
`f['c']` should be `data = f['c']`


kburns:
This line is just internally triggering the change to coefficient space before plotting. The reference you're creating here isn't used -- `data` in the following line is an argument to the lambda function.

PR #248, opened 2023-04-27
Fix _normalize_left_eigenvectors in EigenvalueSolver

### LINKED ISSUE #247: [d3] _normalize_left_eigenvectors
Hi
I've stumbled across a small bug in d3 with the left eigenvalue normalisation. Currently they are normalised as
```
norms = np.diag(modified_left_eigenvectors.T.conj() @ self.eigenvectors)
self.left_eigenvectors /= norms
```
which doesn't account for the conjugate in the inner product. If this normalisation is done then running
```
norms = np.diag(modified_left_eigenvectors.T.conj() @ self.eigenvectors)
```
again after the normalisation does not give a vector of ones for norms.

By changing the second line to
```
self.left_eigenvectors /= np.conj(norms)
```
the left eigenvectors are normalised so a biorthonormal set is produced.
I'm happy to submit a pull request to make this change if it's useful?
Thanks,
Calum

csskene:
This pull request fixes #247 by dividing by the complex conjugate of `norms` rather than by `norms` directly. The resulting modified left eigenvectors then form a bi-orthonormal basis with the right eigenvectors.

### DIFF
--- a/dedalus/core/solvers.py
+++ b/dedalus/core/solvers.py
@@ -160,7 +160,7 @@ class EigenvalueSolver(SolverBase):
     def _normalize_left_eigenvectors(self):
         modified_left_eigenvectors = self._build_modified_left_eigenvectors()
         norms = np.diag(modified_left_eigenvectors.T.conj() @ self.eigenvectors)
-        self.left_eigenvectors /= norms
+        self.left_eigenvectors /= np.conj(norms)
 
     def solve_dense(self, subproblem, rebuild_matrices=False, left=False, normalize_left=True, **kw):
         """


kburns:
@afraser3 any thoughts? If you too are happy with this I'll merge.

afraser3:
Hey @csskene I'm looking more closely at this and it looks like what's already implemented is identical to what was implemented in d2, is that right? I'm pretty sure that in d2 I was able to verify that the existing implementation gives the desired bi-orthonormal basis. Does this mean the same issue exists in d2? Also, any chance you'd be willing to set up a minimal working example to demonstrate this that I could play around with?

csskene:
Hi @afraser3 
You are right that the normalisations are the same, and so the issue does also occur in d2. Attached are examples for both versions 2 and 3 of Dedalus. The examples are the Rayleigh-Bénard EVP examples, modified to also calculate the adjoint eigenvectors. In each case the script prints the normalisations under the current behaviour, and also what the norms would be with the proposed fix.
[examples.zip](https://github.com/DedalusProject/dedalus/files/11422196/examples.zip)

afraser3:
Well this is embarrassing! Calum is completely right, this is a bug in the left_eigenvector normalization in both d2 and d3. I know I was careful in testing this back in the day. I wonder if, because I was originally testing this on an inviscid KH problem which has a certain conjugate symmetry built into the problem, maybe this just wasn't an issue that cropped up for that particular example problem? Or maybe I didn't test as well as I thought I did.

In any case, I agree with Calum's changes. Calum, if you'd be willing, I suggest making the fix in the v2_master branch as well.

csskene:
Hi @afraser3 
Thanks for checking the examples. 
I think that there are a few ways you could have calculated 1 for the normalisation. For example if they are purely real modes you get one either way, and I think the conjugate symmetry you mention can lead to it as well. In my example code if you search for more eigenvalues some of them are real and so you get one in both normalisations. I'm more than happy to make the changes in d2 @kburns @jsoishi ? But I think it needs to be done in an additional pull request?

csskene:
Hi @kburns 
I just had a quick look, and although this issue is fixed in v2 for solve_dense, it is not fixed for solve_sparse. I think all that needs to be done is to apply the same fix to this line
https://github.com/DedalusProject/dedalus/blob/v2_master/dedalus/core/solvers.py#186
as well.

PR #251, opened 2023-06-06
Matsolver updates

kburns:
This PR adds some matsolver updates, updating the interface with umfpack and adding a flexible base class for SuperLU factorized solves that can easily be subclassed to customize all options. Included are NATURAL and COLAMD regular and transposed solvers, since those seem to be the most commonly used.

Relatedly, we should compare NATURAL and COLAMD transposed for all the examples, and see if there's any reason not to switch the default to COLAMD transposed.

@bpbrown @lecoanet @kyle-augustson


kburns:
Merging this in because it's useful. But we should keep in mind the idea of changing the default.

PR #254, opened 2023-06-25
Cleanup tests

kburns:
This PR has a major cleanup of the tests to update to the latest v3 interfaces, add more coverage, and reduce run times.


kburns:
Cleanup isn't finished -- in particular polar, spherical, and transform tests still need cleanup. But there's enough good cleanup here now to merge.

PR #259, opened 2023-07-29
setup.py: support installation when MPI and FFTW dirs have not been specified in environment variables

Kishore96in:
If the user has not set environment variables for the MPI and FFTW directories, the lists `include_dirs` and `library_dirs` contain `None` as an entry. This causes `setup.py` to fail. This PR filters `None` from these lists so that installation can proceed successfully (assuming mpi and fftw are installed in the standard system paths).


kburns:
Thanks!

PR #260, opened 2023-08-02
[d2] Fix left eigenvector normalisation in solve_sparse

csskene:
Similar to pull request #248. This pull request fixes the left eigenvector normalisation in v2_master for the solve_sparse method in the eigenvalue solver. solve_dense has already been fixed in #248.


kburns:
Thanks!

PR #269, opened 2023-09-24
Fully precondition matrices

kburns:
This PR takes advantage of the fact that our preconditioners are now semi-orthonormal, so their pseudoinverses are trivial to compute and apply. This means we no longer need to store non-right-preconditioned versions of the matrices for dot products during timestepping. We can therefore just store fully preconditioned versions, and apply the various preconditioners during the gather and scatter calls for each subproblem. This really simplifies the timestepping and other solver code.

The PR also includes some changes to reduce allocations during the timestepping loop. In particular, views into the field variables are cached for each subsystem to speed up the gather/scatter steps. The only allocations in the timestepping, of either buffers or new array views, should just be inside the F evaluation or by the matsolver itself.

### DIFF
--- a/dedalus/core/evaluator.py
+++ b/dedalus/core/evaluator.py
@@ -148,28 +148,35 @@ class Evaluator:
 
     def require_coeff_space(self, fields):
         """Move all fields to coefficient layout."""
+        coeff_layout = self.dist.coeff_layout
+        # Quickly return if all fields are already in coeff layout
+        if all(f.layout is coeff_layout for f in fields):
+            return
         # Build dictionary of starting layout indices
-        layouts = defaultdict(list, {0: []})
+        layouts = defaultdict(list)
         for f in fields:
-            layouts[f.layout.index].append(f)
-        # Decrement all fields down to layout 0
-        max_index = max(layouts.keys())
+            if f.layout is not coeff_layout:
+                layouts[f.layout.index].append(f)
+        # Decrement all fields down to coeff layout
         current_fields = []
-        for index in range(max_index, 0, -1):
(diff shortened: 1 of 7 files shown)


kburns:
@lecoanet any hesitation to merge?

jsoishi:
This looks _really_ good to me!

lecoanet:
Have we done any performance testing of this at scale? That is my hesitation. I feel like we have tested lots of things that seem like they should make things faster, but actually slowed things down, so it would be good to know. But I don't have a good performance test. I guess we could use a higher-resolution version of some of the 3D test problems?

kburns:
On 64 cores on Flatiron, here are the runtimes of the internally heat convection example in the ball. Resolution is (128, 64, 96) and its about 1000 iterations.

master: 182.6s
new: 180.9s

lecoanet:
I tested two problems on 2048 Pleiades ivybridge cores:

3D cartesian convection/stably stratified simulation, Nx=Ny=256, matched Chebyshev with Nz=512+256. Ran for 100 iterations (after 10 iteration startup). I ran a couple of times and found the timing appeared to be reproducible.

master: 66.2 sec
PR: 66.8 sec

3D shell example problem, Nphi=1024, Ntheta=512, Nr=512. Ran for 100 iterations (after 10 iteration startup). Here are the run times running it twice:

master: 197.0, 197.1
PR: 195.7, 197.0

So for larger problems, it seems like it maybe hurts/helps by ~1% of the run time, so I'm happy with the performance.

kburns:
I've moved the head of master back to 5a8bbaf because the merge here included updates to cython which subsequently quickly broke. Thats now a separate branch which I'll rebase and merge to this when cython is fixed.

PR #271, opened 2023-10-18
Change default matsolvers to colamd

kburns:
The COLAMD ordering seems to be much better at reducing fill-in than NATURAL, particularly for multi-dim. coupled problems. I'm not aware of cases with substantial performance degradations, but haven't looked thoroughly yet...

### DIFF
--- a/dedalus/dedalus.cfg
+++ b/dedalus/dedalus.cfg
@@ -87,10 +87,10 @@
 [linear algebra]
 
     # Default sparse matrix solver for single solves
-    MATRIX_SOLVER = SuperLUNaturalSpsolve
+    MATRIX_SOLVER = SuperLUColamdSpsolve
 
     # Default sparse matrix factorizer for repeated solves
-    MATRIX_FACTORIZER = SuperLUNaturalFactorizedTranspose
+    MATRIX_FACTORIZER = SuperLUColamdFactorizedTranspose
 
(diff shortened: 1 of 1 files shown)


lecoanet (approved):
I tested two problems on 2048 Pleiades ivybridge cores:

3D cartesian convection/stably stratified simulation, Nx=Ny=256, matched Chebyshev with Nz=512+256. Ran for 100 iterations (after 10 iteration startup).

master: 66 sec (34 sec startup)
PR: 55 sec (17 sec startup)

I also used the memtowel script to see how much memory is being used. On master, the highest memory usage on one core was ~1.2GB, but on PR it was ~0.5GB.

3D shell example problem, Nphi=1024, Ntheta=512, Nr=512. Ran for 100 iterations (after 10 iteration startup). Here are the run times running it twice:

master: 203.5, 204.1 (startup time of 51, 52 sec)
PR: 203.7, 204.5 (startup time of 64, 52 sec)

The differences from the PR are within the run-to-run variability. It makes sense the effect is smaller for the shell problem because the matrix solves comprise a smaller fraction of the run time.

Also, the memory usage does not change significantly due to the PR for the shell problem (max of ~1.4GB per core in both cases).

From these problems, it seems like the PR can give significant speedups for Cartesian problems, but maybe does not change things much for shell problems. Seems like it is a good idea to change this default solver.

PR #275, opened 2023-12-21
DirectProduct bases

kburns:
This PR adds support for direct products between bases with separable intertwiners (Cartesian and spin bases). This has primarily been implemented for and tested on axially periodic cylinders.


kburns:
Tests passing (except skipped NCC tests)

lecoanet:
Cylinder NCC tests sometimes fail due to transform noise. In the future we should compare the coefficient data so the tests are really tests of only the NCCs. Right now they're both tests of NCCs & transforms.

lecoanet (approved):

jsoishi (approved):

PR #276, opened 2023-12-22
Linalg improvements

kburns:
This PR includes some improvements to linear algebra using cython routines.


kburns:
For 2D RBC at (Nx, Nz) = (256, 512), this speeds things up by about 5%, and for 3D HOT at 256^3 it speeds things up by about 4%. Bigger differences on other more anisotropic simulations where inverse Jacobi added a lot.

kburns:
The old behavior, if desired, is available through the OLD_CSR_MATVECS config option.

PR #277, opened 2024-01-05
Reuse the LU transform when solving for left eigenvectors using the sparse solver

csskene:
SuperLU supports solving the transposed system from the LU factorisation of the un-transposed system. This PR adds a *solve_adjoint* method to the SuperLU matrix solvers so that the LU factorisation from the right eigenvector solve can be reused for the left eigenvector solve. Corresponding changes are made to the sparse evp solver. 

The benefits are that only one LU factorisation now needs to be done for each sparse solve which improves efficiency as this is often the slowest part of the evp solve. Furthermore, as matrix factorisations are based on the direct matrices, factorising the transposed system is often slower due to its change in structure. This PR removes this issue.

### DIFF
--- a/dedalus/libraries/matsolvers.py
+++ b/dedalus/libraries/matsolvers.py
@@ -136,13 +136,8 @@ class _SuperluFactorizedBase(SparseSolver):
     def __init__(self, matrix, solver=None):
         if self.trans == "T":
             matrix = matrix.T
-            self.trans_H = "N"
         elif self.trans == "H":
             matrix = matrix.H
-            self.trans_H = "N"
-        else:
-            self.trans_H = "H"
-
         self.LU = spla.splu(matrix.tocsc(),
                             permc_spec=self.permc_spec,
                             diag_pivot_thresh=self.diag_pivot_thresh,
@@ -152,12 +147,14 @@ class _SuperluFactorizedBase(SparseSolver):
 
     def solve(self, vector):
         return self.LU.solve(vector, trans=self.trans)
-    
+
     def solve_H(self,vector):
-        if(self.trans=="T"):
-            return np.conj(self.LU.solve(np.conj(vector), trans=self.trans_H))
-        else:
-            return self.LU.solve(vector, trans=self.trans_H)
+        if self.trans == "N":
+            return self.LU.solve(vector, trans="H")
+        elif self.trans == "H":
+            return self.LU.solve(vector)
+        elif self.trans == "T":
+            return np.conj(self.LU.solve(np.conj(vector)))
 
 
 @add_solver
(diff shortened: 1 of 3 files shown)


dedalus/tools/array.py
```diff
@@ -427,7 +429,24 @@ def matvec(x):
     evals, evecs = spla.eigs(D, k=N, which='LM', sigma=None, **kw)
     # Rectify eigenvalues
     evals = 1 / evals + target
-    return evals, evecs
+    
+    if(left==True):
```
kburns, line 433: Should just be `if left:`

dedalus/tools/array.py
```diff
@@ -427,7 +429,24 @@ def matvec(x):
...
+        # Build sparse linear operator representing (A^H - σB^H)^I B^H = C^-H B^H = D_adj
```
kburns, line 434: Should this be conj(sigma)?

dedalus/tools/array.py
```diff
@@ -427,7 +429,24 @@ def matvec(x):
...
+        # Note: D_adj is not equal to D^H
+
```
kburns, line 436: Please remove extra blank lines

csskene:
Thanks for the review, I've tried to address the comments with the latest push.

dedalus/tools/array.py
```diff
@@ -427,7 +429,19 @@ def matvec(x):
...
     evals = 1 / evals + target
-    return evals, evecs
+    if left:
+        # Build sparse linear operator representing (A^H - conj(σ)B^H)^I B^H = C^-H B^H = D_adj
+        # Note: D_adj is not equal to D^H
+        def matvec_adj(x):
+            return solver.solve_adjoint(B.H.dot(x))
+        D_adj = spla.LinearOperator(dtype=A.dtype, shape=A.shape, matvec=matvec_adj)
+        # Solve using scipy sparse algorithm
+        evals_adjoint, evecs_adjoint = spla.eigs(D_adj, k=N, which='LM', sigma=None, **kw)
```
kburns, line 439: I think we should call these `left_` instead of `_adjoint` throughout to be consistent with scipy.

dedalus/libraries/matsolvers.py
```diff
@@ -24,6 +24,9 @@ def __init__(self, matrix, solver=None):
     def solve(self, vector):
         pass
 
+    def solve_adjoint(self,vector):
```
kburns, line 27: Maybe we should all this `solve_H` instead of `solve_adjoint`, to match the interface of superlu and be a little more explicit, in case we/someone wants `solve_T` at some point.

dedalus/tools/array.py
csskene: Agreed, have updated the code.

dedalus/libraries/matsolvers.py
csskene: Also agreed, have changed this in the code too.

kburns:
Thanks! Looks good, I might push a quick change to cleanup the matrix solve stuff a bit, then merge.

PR #280, opened 2024-01-24
Half dimension docs

kburns:
This PR adds a doc page describing how to setup coordinate systems and processes for even distribution of half dimensional problems.


kburns:
Merging now to include in the next bug-fix release.

PR #288, opened 2024-03-14
Profiling enhancements

kburns:


kburns:
This adds some extra wrapper layers around the cython routines to make them appear in cprofile. Some of this will hopefully be supported again in cython3 and python 3.12+, but we don't know how soon that will be.

PR #290, opened 2024-04-29
Modernize python build configuration

MilesCranmer:


setup.py
```diff
@@ -217,24 +184,38 @@ def run(self):
...
+
+    # Write the version to version.py
+    with open(os.path.join(os.path.dirname(__file__), "dedalus", "version.py"), "w") as f:
+        f.write(f'__version__ = "{version}"')
+
+    kwargs = {
+        "use_scm_version": False,
+        "version": version,
+    }
+
```
MilesCranmer, line 214: (The same one I use for PySR, so I can confirm this works!)

MilesCranmer:
Friendly ping @kburns

requirements.txt
```diff
@@ -0,0 +1,12 @@
+docopt
```
VascoSch92, line 1: You can also move te dependencies to the project.toml.

Moreover, you can divide them between the ones used but the package and the ones used for dev ;)

It is a file less

MilesCranmer: I prefer them separate for backwards compatibility since many workflows involve a `pip install -r requirements.txt` step. The `dynamic` keyword in the pyproject enables this. 

Also I’m a bit confused, are you a dedalus developer or just passing by? It says you reviewed this PR but GitHub doesn’t label you a “member” or “contributor”.

(I’m seeking the preferences of the dedalus core team here if that is ok)

VascoSch92: As it is open source everyone can comment.


It was just a suggestion 😉

If you would like the review from a mantainer/core dev, you can directly ask it.

MilesCranmer:
Ping @kburns

PR #294, opened 2024-06-18
Fix dispatching issues with get_basis on cartesian coordinates

### LINKED ISSUE #293: CartesianGradient / Laplacian bug
There seems to be a bug in CartesianGradient which was found by Stephane Labrosse & detailed here:

https://groups.google.com/u/2/g/dedalus-users/c/UcOEEzv25TI

He has a compressible convection script that, when run for several thermal times, should reach thermal equilibrium with the same Nusselt number at top and bottom boundary, and with constant flux. This worked correctly up through commit ff3aea3, then the following commit 1218c62 which modified the CartesianGradient operator breaks the script and causes the simulation to generate NaNs within 10 timesteps. The later commit f263415 further modifies the CarteisanGradient operator -- at this commit, the script runs but does not generate correct results, e.g., the Nusselt number at top and bottom are different, and the total flux is not constant.

kburns:
Fixes #293

### DIFF
--- a/dedalus/core/operators.py
+++ b/dedalus/core/operators.py
@@ -3415,8 +3415,8 @@ class CartesianDivergence(Divergence):
     @classmethod
     def _preprocess_args(cls, operand, index=0, out=None):
         coordsys = operand.tensorsig[index]
-        if operand.domain.get_basis(coordsys) is None:
-            raise SkipDispatchException(output=0)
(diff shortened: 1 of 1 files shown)


lecoanet (commented):
Tested Stephane Labrosse's script, and this fixes his issue.

PR #307, opened 2024-10-04
Adaptive stepper

danvau98:
Additional IMEX scheme with an embedded scheme to allow for adaptive stepping.


danvau98:
I want to find some lower-order embedded methods that would increase the speed of these solvers. The Kennedy and Carpenter schemes are  4th and 5th-order; ideally, I should find a 3rd and 2nd-order scheme to add.

danvau98:
Hi @bpbrown and @kburns, any suggestions for how this can be improved?

dedalus/core/timesteppers.py
```diff
@@ -727,3 +731,282 @@ class RKGFY(RungeKuttaIMEX):
...
+                # Slice out valid subdata, skipping invalid components
+                spRHS = RHS.get_subdata(sp)
+                spX = sp.LHS_solvers[i].solve(spRHS)  # CREATES TEMPORARY
+                sp.scatter_inputs(spX, state_fields)
+
+            # If this is the last iteration, swap H's last row with b_hat and compute X_hat
+            if i == self.stages-1:
+                # Swap the last row of H with b_hat
+                H[-1, :] = b_hat_IM
+                # A[-1, :] = b_hat_EX <- not sure if this is needed
```
brownbaerchen, line 891: This is interesting. I have a few questions:
 - Did you check that you get the desired order of accuracy for both methods? I understand that you cannot just sum because of algebraic constraints. But if you solve, you are doing something different. If you get the correct order, is it by chance or is it expected?
 - Why did you comment out the above line? I would expect that it is needed in general. 
 - How many more factorisations do you need with this scheme? The RK methods so far implemented in Dedalus are all singly diagonally implicit, such that you need only one factorisation per step. But I assume $\hat{b}_s \neq H_{ss}$, such that you need one more factorisation. If the step size is kept constant, do you need to factorize twice again in the next step or is there a cache?

PR #317, opened 2025-03-19
Following commit a1d7884 on master branch, updates sparse .A to .toarray() as the former is deprecated in scipy 1.14

afraser3:
I don't see other related issues in v2 besides this one line, but I fear I may have missed something.


afraser3:
I wonder if [this line](https://github.com/DedalusProject/dedalus/blob/v2_master/dedalus/libraries/matsolvers.py#L187) also needs the same change. Admittedly I'm a bit out of my element here.

PR #318, opened 2025-04-16
Angular momentum operators

kburns:


dedalus/core/operators.py
```diff
@@ -3975,6 +3913,300 @@ def operate(self, out):
...
+
+    @CachedMethod
+    def radial_matrix(self, regindex_in, regindex_out, ell):
+        radial_basis = self.radial_basis
+        regtotal_in = radial_basis.regtotal(regindex_in)
+        regtotal_out = radial_basis.regtotal(regindex_out)
+        if regindex_in[1:] == regindex_out[1:]:
+            return self._radial_matrix(radial_basis, regindex_in[0], regindex_out[0], regtotal_in, regtotal_out, ell)
+        else:
+            raise ValueError("This should never happen")
```
VascoSch92, line 3965: If I can suggest something: in the error message give some more information why this has happened. Otherwise, it is not so helpful... For example

```
ValueError(f"This should never happen, got {refined_in=}")
```

dedalus/core/operators.py
```diff
@@ -3975,6 +3913,300 @@ def operate(self, out):
...
+        if regindex_in0 == Rm and regindex_out0 == R0:
+            return -1j * radial_basis.xi(+1, ell+regtotal_in+1) * radial_basis.operator_matrix('R+', ell, regtotal_in)
+        elif regindex_in0 == Rp and regindex_out0 == R0:
+            return 1j * radial_basis.xi(-1, ell+regtotal_in-1) * radial_basis.operator_matrix('R-', ell, regtotal_in)
+        elif regindex_in0 == R0 and regindex_out0 == Rm:
+            return -1j * radial_basis.xi(+1, ell+regtotal_in) * radial_basis.operator_matrix('R-', ell, regtotal_in)
+        elif regindex_in0 == R0 and regindex_out0 == Rp:
+            return 1j * radial_basis.xi(-1, ell+regtotal_in) * radial_basis.operator_matrix('R+', ell, regtotal_in)
+        else:
+            raise ValueError("This should never happen")
```
VascoSch92, line 4071: same as before :-)

PR #331, opened 2026-02-24
Optimize build_cartesian_ncc_matrix

tjc0726:
This PR optimizes `Product.build_cartesian_ncc_matrix()` in `dedalus/core/arithmetic.py` to achieve speedup for problems with spatially varying coefficients (NCC).

The key insight is to batch-accumulate sparse matrices instead of sequential addition, reducing sorting/deduplication overhead from O(n²) to O(n log n).

In my case

Benchmark: 2D Navier-Stokes with NCC, N=64
- **Original**: 297.63s per Newton iteration
- **Optimized**: 27.23s per Newton iteration
- **Speedup**: 10.93x

Numerical Verification
All fields match to machine precision (relative error < 1e-14):
- u: 3.789e-14
- p: 2.028e-12
- force_x: 6.609e-16

here's my example using monkry patch, one can delete the patch to check the difference

[example_ncc_flow_optimized.py](https://github.com/user-attachments/files/25503563/example_ncc_flow_optimized.py)

### DIFF
--- a/dedalus/core/arithmetic.py
+++ b/dedalus/core/arithmetic.py
@@ -430,18 +430,31 @@ class Product(Future):
             Gamma = Gamma.transpose((2, 0, 1))
         # Loop over NCC modes
         shape = (subproblem.field_size(out), subproblem.field_size(arg))
-        matrix = sparse.csr_matrix(shape, dtype=self.dtype)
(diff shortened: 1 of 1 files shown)


kburns:
Thanks! This is a nice approach and I'll go through and optimize things a bit more to use just the COO data internally.

PR #332, opened 2026-02-24
Cardinal basis

kburns:
This PR adds a CardinalBasis class, which is a 1D basis of independent points with no differentiability. This may be useful for doing simulations with parameter dimensions, etc.  The only supported operators are ConvertConstant, Interpolate (to extract a particular index), Integrate (discrete sum over entries), and Average (average over entries).  I have tested it with an implementation of 2D RBC for running separate initial conditions simultaneously, but haven't put in any unit tests.


jsoishi:
This looks great!

